PR #10, opened 2020-05-08
Add reactions

### LINKED ISSUE #8: Add Reactions
This is similar to #7 

Add reactions of the type:
- like
- dislike
- flag/report

To add these things, I think we would need an abstract model to extend them(reactions) as property. Any other ideas would be great.

Before adding these fields, we need to answer a couple of questions:
- Would we want these reactions to be used only if a user decides to use them? Preferably a configuration in the settings file, that allows them to add these fields.
- What could be the potential reasons for flagging/reporting?
- How would the flagging process work? 
     - I had made an application for flagging which is somewhat usable right now. Would the process be similar to this one? 
here

I would be willing to put together a pull request if we can decide about the details regarding the implementation.

abhiabhi94:
Add reactions to comments.

Tasks:
- [x] Add models for Reaction and ReactionInstance
    - [x] Add logic for the models.
- [x] Make changes in the comments model
    - [x] Add a `post_save` signal that creates a Reaction instance when a comment is created. 
    - [x] Add likes and dislikes count as a property.
- [x] Add urls and views
    - [x] For django.
    - [x] For REST API.
- [x] Add template tag to return if the current user has shown a particular reaction(like/dislike) on a particular comment.
- [x]  Fix Frontend
    - [x] Display reaction count on the frontend using icons.
    - [x] Perform Reaction using AJAX request for authenticated users.
    - [x] Redirect unauthenticated users to login page with correct path for the next redirect.
- [x]  Add unit tests for
    - [x] models.
    - [x] views.
    - [x] REST API.
    - [x] template tags.
- Fix leftovers
    - [x] Add additional tests to increase coverage.

The reaction process currently looks something like this:
comment-reactions

Fixes #8, maybe we can move adding flag to a separate issue.


comment/admin.py
```diff
@@ -1,5 +1,6 @@
 from django.contrib import admin
-from comment.models import Comment
+
+from comment.models.comments import Comment
```
Radi85, line 3: Since we imported the models in `__init__` we can terminate the module name from the import statement.
`from comment.models import Comment`  <-- I prefer this form.

comment/models/reactions.py
```diff
@@ -0,0 +1,90 @@
+from collections import namedtuple
+from typing import List
```
Radi85, line 2: Please don't use `typing`

comment/models/reactions.py
```diff
@@ -0,0 +1,90 @@
...
+
+# construct a named tuple
+choices = namedtuple('reaction_choices', ['value', 'reaction'])
+
+# construct the list of named tuple
+reaction_choices = [choices(*reaction) for reaction in REACTION_CHOICES]
+
+
+class Reaction(models.Model):
+    reactions = models.SmallIntegerField(choices=REACTION_CHOICES, default=reaction_choices[0].value)
```
Radi85, line 26: In this model design, a new row in the db will be created for each new reaction. More efficient solution could be as follows:
``` python
from django.db import models

class Reaction(models.Model):
    comment = models.ForeignKey(Comment, related_name='comment_reaction', on_delete=models.CASCADE)
    like = models.PositiveIntegerField(default=0)
    dislike = models.PositiveIntegerField(default=0)
    
    objects = models.Manager()
    comment_objects = ReactionManager()

    def increase_like(self):
        self.like = models.F('like') + 1
        self.save()
        
    def increase_dislike(self):
        self.dislike = models.F('dislike') + 1
        self.save()

    def decrease_like(self):
        if self.like > 0:
            self.like = models.F('like') - 1
            self.save()
            
    def decrease_dislike(self):
        if self.dislike > 0:
            self.dislike = models.F('dislike') - 1
            self.save()
      
```

I believe we could get better performance on this model since only one reaction entry will be created for each comment.

Radi85 (commented):
Hey,

Many thanks for you efforts.

I would separate `show reacted user` feature in a different issue so for now just ignore it.

Please consider PEP 8 compatibility and ignore typing for now.

comment/models/reactions.py
abhiabhi94: Hey @Radi85 , by this design we don't associate a user to a reaction. This will raise a couple of issues:
- One user can like/dislike a comment multiple times.
 - We wouldn't be able to extend this (maintaining backward compatibilities) to show users if we want it in the future.

comment/models/reactions.py
abhiabhi94: Any particular reason? They were introduced in 3.5 and we support versions later to their release. I like them personally but don't have an issue if you want them to be removed.

comment/models/reactions.py
Radi85: Hey,
The user has nothing to do with this model because the user is still connected with the other mode `ReactionInstance`.
to keep traking of the user rection type we could add one more field `reaction_type` to `ReactionInstance` model.
The user will be able at anytime to like, dislike or remove his action.

comment/models/reactions.py
Radi85: It's just a personal openion, I don't like typing with python. However, in the future we may open an issue to refactor the code and add them.

comment/models/reactions.py
abhiabhi94: Okay, I misunderstood it. I thought you were asking to remove the `ReactionInstance` model. So I had used the reactions field so that we could extend the model in the future if we want more reactions like love, confused, etc. So, do you still want me to go through with your suggestion?

Radi85: Yes please. Adding new columns for new reactions is much more cheaper than creating new entry in the db for each new reaction.

Radi85 (changes requested):

comment/admin.py
```diff
@@ -1,5 +1,7 @@
 from django.contrib import admin
-from comment.models import Comment
+
+from comment.models import Comment, Reaction
+from comment.models.reactions import ReactionInstance
```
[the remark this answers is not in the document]
abhiabhi94: Earlier all `models` were not imported inside the `__init__.py` file. I will import everything inside that file and make this change.

comment/migrations/0004_reaction_reactioninstance.py
```diff
@@ -0,0 +1,38 @@
...
+                ('reaction_type', models.SmallIntegerField(choices=[(1, 'Like'), (2, 'Dislike')], default=1)),
+                ('date_reacted', models.DateTimeField(auto_now=True)),
+                ('reaction', models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, related_name='reaction', to='comment.Reaction')),
+                ('user', models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, related_name='user', to=settings.AUTH_USER_MODEL)),
+            ],
+            options={
+                'unique_together': {('user', 'reaction')},
+            },
+        ),
+    ]
```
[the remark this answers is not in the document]
abhiabhi94: I have added only 1 migration file as of now.

comment/models/reactions.py
```diff
@@ -0,0 +1,125 @@
...
+
+        Returns:
+            None
+        """
+        if reaction == ReactionInstance.ReactionType['LIKE']:
+            self.decrease_likes()
+        else:
+            self.decrease_dislikes()
+
+    def update_reaction(self, user, comment, reaction):
```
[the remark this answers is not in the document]
abhiabhi94: I had made the other functions private as of now. Hence I added the function to this model.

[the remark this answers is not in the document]
Radi85: It's your call. Just my openion to keep the code cleaner.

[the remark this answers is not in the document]
abhiabhi94: Hey, I'm from Uttar Pradesh, a state in India. We are not used to such types of empowering choices :-p
Jokes apart I have moved the function to the `ReactInstance` model and refactored it. Please have a look and tell me if it's okay.

comment/models/reactions.py
```diff
@@ -0,0 +1,199 @@
...
+class Reaction(models.Model):
+
+    comment = models.ForeignKey(Comment, related_name='reactions', on_delete=models.CASCADE)
+    likes = models.PositiveIntegerField(default=0)
+    dislikes = models.PositiveIntegerField(default=0)
+    
+    objects = models.Manager()
+    comment_objects = ReactionManager()
+    
+    def _increase_likes(self):
```
Radi85, line 25: In my openion all this model's methods are not private since all of them will be called outside the class.

comment/models/comments.py
```diff
@@ -0,0 +1,59 @@
...
+    @property
+    def replies(self):
+        return Comment.objects.filter(parent=self).order_by('posted')
+
+    @property
+    def is_edited(self):
+        return self.posted.timestamp() + 1 < self.edited.timestamp()
+
+    @property
+    def likes(self):
```
Radi85, line 54: In fact we don't need this property since we can get the reaction count straight forward:
```python
comment = Comment()
comment.reaction.likes
comment.raction.dislikes
```
Keeping them doesn't hurt.

comment/api/urls.py
```diff
@@ -1,11 +1,14 @@
...
 from rest_framework.urlpatterns import format_suffix_patterns
+
 from comment.api import views
+from comment.views import react
 
 urlpatterns = [
     path('comments/', views.CommentList.as_view(), name='comments-list'),
     path('comments/create/', views.CommentCreate.as_view(), name='comments-create'),
     path('comments/<int:pk>/', views.CommentDetail.as_view(), name='comment-detail'),
+    path('comments/react/<int:comment_id>/<str:reaction>', react, name='react')
```
Radi85, line 11: For more conventions could you please modify the url as follows:
```python
path('comments/<int:comment_id>/react/<str:reaction>/', react, name='react')
```

Radi85 (changes requested):
hey,
Thanks a lot. A lot of work has been done.
I would like to use icons instead, thumbs up and down, if your question is realted to the font, feel free to use any light weight version.
As suggested, please move `reply`, `like` and `dislike` to the second row.
Unfortunately I have no more time but I would like to get back to review the models again, nice work!

Radi85:
Btw, the work you have already done, could be splitted to more features :D
I will have a look at the tasks a bit later.

comment/api/urls.py
abhiabhi94: okay. In the other `url` file that renders template views, we use the `app_name` as `comment` and here as `comments`. Is this expected? Also, there this convention has not been followed. Should I make the change there too or leave it for backward compatibility?

comment/models/comments.py
abhiabhi94: I have added these property so that these can be used directly in the templates without defining any other template tag or functions to send in the count. Just noticed the typo, will correct it too.

comment/api/urls.py
Radi85: Also we need to change the `react` url in the other file to look like my suggestion here. `edit` and `delete` are actions performed on the comment model directly while `react` comes from another model, therefore, I would use comment id right after `comments/` in case of react or any other action not related to the comment model directly.

comment/views/reactions.py
```diff
@@ -0,0 +1,35 @@
...
+from django.http import HttpResponseBadRequest, JsonResponse
+from django.shortcuts import get_object_or_404
+from django.utils.translation import gettext_lazy as _
+from django.views.decorators.http import require_POST
+
+from comment.models import Comment, ReactionInstance
+
+@require_POST
+@login_required
+def react(request, comment_id, reaction):
```
Radi85, line 11: Can you please use CBV instead of FBC? CBV is more clean choice.

comment/models/reactions.py
```diff
@@ -0,0 +1,199 @@
...
+            reaction_type (int): The integral value that matches to the reaction value in
+                the database
+        """
+        super().save(*args, **kwargs)
+        reaction = self.reaction
+        reaction_type = self.reaction_type
+        reaction._increase_reaction_count(reaction_type)
+
+    @classmethod
+    def _delete_and_create_new_instance(cls, instance, user, reaction_type):
```
Radi85, line 132: What the purpose of this method?
Can we adopt the following workflow:
* Whenever a new comment is created, a new Reaction entry will be created as well (use signals).
* When the user reacts on a comment for the first time, the ReactionInstance entry will be created. This entry will live as long as both the user and the comment are there, user or comment is deleted, the entry will be deleted.
* Further reactions from the user:

    - `like` => in Reaction model `increate_likes` called and in ReactionInstance `reaction_type` is set to 1
    - `dislike` same above with corresponding values.
    - `unlike` or `undislike` When the user press again on the liked or disliked icon the corresponding reaction should be decreased and the `reaction_type` in the ReactionInstance should be set to `(0, NEUTRAL)`

comment/models/reactions.py
```diff
@@ -0,0 +1,199 @@
...
+        if reaction == ReactionInstance.ReactionType.LIKE.value:
+            self._decrease_likes()
+        else:
+            self._decrease_dislikes()
+
+
+class ReactionInstance(models.Model):
+    
+    @unique
+    class ReactionType(IntEnum):
```
Radi85, line 83: This will be required as mentioned below.
```python
class ReactionType(IntEnum):
    NEUTRAL = 0
    LIKE = 1
    DISLIKE = 2
```

Radi85 (changes requested):

comment/models/reactions.py
abhiabhi94: This function deletes a previous instance of `ReactInstance` that the user had and creates a new one if required.
Use cases
- The user has already liked a comment and they again make a request to like the comment, it will only delete the previous instance. 
-  The user has already liked a comment and they make a request to dislike the comment, tit will delete the previous instance and create a new instance for dislike.




>     * `unlike` or `undislike` When the user press again on the liked or disliked icon the corresponding reaction should be decreased and the `reaction_type` in the ReactionInstance should be set to `(0, NEUTRAL)`

Why do we need a third reaction? We can just delete the previous instance. The user could have mistakenly clicked on the like icon. In case you want it just as a default value for the database, is it necessary in our case?

comment/views/reactions.py
abhiabhi94: Do we really need a class based view? This function currently handles pretty basic stuff.

comment/models/reactions.py
Radi85: The instance is actually a db entry, my suggestion is to avoid deleting it. This will give us idea of the user activity. NUTRAL type will show us that the use has reacted with a comment.

comment/views/reactions.py
Radi85: Since version 1.4.0 all FBVs have been converted to CBVs. You can leave it as is and I may refactore it later.

Radi85:
Well done so far. I like how it looks like.

comment/models/reactions.py
abhiabhi94: Yes I do get your point but the user could also mistakenly click on one of the reaction. It would not always be a correct measure of their reaction. It could be useful if the reactions weren't mutually exclusive.

comment/templates/comment/content.html
```diff
@@ -7,9 +7,9 @@
         <small>
             <a class="{% block user_name %}{% endblock user_name %}" href="{% get_profile_url obj %}">{{ obj.user }}</a> |
             {% if obj.is_edited %}
-            <cite title="Edited">Edited: {{ obj.edited|timesince }} ago</cite>
+            <span class="font-italic" title="Edited">Edited: {{ obj.edited|timesince }} ago</span>
```
abhiabhi94, line 10: I believe the `cite` element was here mistakenly. As far I can see [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/cite) , I do not think we needed it here.

abhiabhi94:
Hey, currently the reaction thing works well for authenticated users. 

For unauthenticated users, I was thinking of following this workflow
- Inside the template,  add an attribute to the anchor element and set the `href` attribute accordingly. Then inside the javascript, when the `onclick` event is triggered, I could check if the attribute is set, redirect them to the login page. 
    - An issue here is the `next` path to the login URL in `django` redirects through `GET` requests only?(I may be wrong here). Currently, I had allowed only `POST` requests for accepting reactions. Should I change the request method there?

- How should we handle `oauth`?

comment/models/reactions.py
abhiabhi94: Only `increase_reaction_count` and `decrease_reaction_count` were called from outside the class. I have made them public now.

Radi85:
> For unauthenticated users, I was thinking of following this workflow

To avoid this complex, please use a simple if statement inside the template:
```jinja2
{% if user.is_authenticated %}
 a_tag
{% else %}
div_tag
{% endif %}
```

> * How should we handle `oauth`?

I believe django will authenticate the user when using `oauth`, so you have nothing to do here.

comment/templates/comment/content.html
```diff
@@ -1,14 +1,15 @@
 {% load comment_tags %}
+{% load i18n %}
```
Radi85, line 2: what the purpose of loading i18n?

comment/templatetags/comment_tags.py
```diff
@@ -139,3 +140,48 @@ def include_bootstrap():
...
+        arg2 (`Any`): Second Argument
+
+    Returns:
+        tuple: a tuple containing both the arguments in the order that they were passed.
+    """
+    return arg1, arg2
+
+
+@register.filter(name='has_reacted')
+def has_reacted(comment_and_user, reaction):
```
Radi85, line 161: Sorry for the late review, I saw that you have written the test already but it is really worthy to refactor the code here.
Instead of using these two filter, we can use a simple tag directly as follows:
```python
@register.simple_tag(name='has_reacted')
def has_reacted(comment, user, reaction):
    if user.is_authenticated:
        reaction_type = getattr(ReactionInstance.ReactionType, reaction.upper(), None)
        if not reaction_type:
            raise template.TemplateSyntaxError(
                'Reaction must be an valid ReactionManager.RelationType. {} is not'.format(reaction)
                )
        return ReactionInstance.objects.filter(
            user=user,
            reaction_type=reaction_type.value,
            reaction__comment=comment
            ).exists()

    return False

```

Use in temple:
```jinja2
{% if hast_reacted comment=obj user=user reaction="like" %}"#ffc966"{% else %}"none"{% endif %}
```

or if you like you can return the color directly instead of the boolean value.

Radi85 (changes requested):

abhiabhi94:
> To avoid this complex, please use a simple if statement inside the template:

How will this help in handling the event when triggered through javascript?

 > I believe django will authenticate the user when using oauth, so you have nothing to do here.

I was concerned about redirect the user to the like link for the comment. How do we do that?

comment/templates/comment/content.html
abhiabhi94: It was added previously when we had the text(`Like`) in place of icons. I will use a title attribute to the images and put the text inside the `trans` tag. Just to reduce our work when we decide to make the application to be portable to other languages than English.

comment/templatetags/comment_tags.py
abhiabhi94: Okay, I will do that. Returning the color from the template tag function would make it a little too complex for the users to change the icon color if they want. I will keep the color inside the template.

Radi85:
> How will this help in handling the event when triggered through javascript?

Javascript fucntion is triggered here:
```javascript
 $(document).on("click", ".js-comment-reaction", commentReact);
```
Add this class to html tag only if the user is authenticated:

```jinja2
<small class="{% if user.is_authenticated %}js-comment-like {% endif %} {% block like_link_cls %} btn btn-link" {% endblock like_link_cls %} href="{% url 'comment:react' comment_id=obj.id reaction='like' %}">{% trans "Like" %}</small>
```

> I was concerned about redirecting the user to the like link for the comment. How do we do that?

I am not sure what `redirecting the user to the like link for the comment` means. Our app is associated with a given model and our urls are related to that model's urls. I believe this redirecting part shall be handeled by the developer who uses our app.

abhiabhi94:
> I am not sure what redirecting the user to the like link for the comment means. Our app is associated with a given model and our urls are related to that model's urls. I believe this redirecting part shall be handeled by the developer who uses our app

Sorry I didn't mention this explicitly, but I was talking about unauthenticated users. For normal auth we can put the link something like this `/login?next=redirect_link` but what to do in this case

abhiabhi94:
We also need to add exception handling to the comment views.
Testing your comment views I found that exception handling has not been handled if the URLs are accessed directly through the browser without form attributes or other values.

Radi85:
> We also need to add exception handling to the comment views.
> Testing your comment views I found that exception handling has not been handled if the URLs are accessed directly through the browser without form attributes or other values.

You are right. It's better to open a separate issue for this.

abhiabhi94:
Hey, why did the test fail? I think I have not changed anything that should cause this.

Radi85:
> Hey, why did the test fail? I think I have not changed anything that should cause this.

The test is passed for some enironment and fails on others, Traceback:

```
FAIL: test_create_comment (comment.tests.test_views.CreateCommentTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/Radi85/Comment/comment/tests/test_views.py", line 45, in test_create_comment
    self.assertIsNone(response.context.get('comment'))
AssertionError: reply by test-1: child comment body is not None
```

abhiabhi94:
Could you guide me with what documentation needs to change and what should be added?. Apart from that, you may check and tell me if anything else needs to be done. In the meantime, I will try to figure out and increase test coverage.

comment/models/__init__.py
```diff
@@ -0,0 +1,3 @@
+# flake8: noqa
+from comment.models.comments import *
+from comment.models.reactions import *
```
Radi85, line 3: Please add empty line in the end of file.

comment/templates/comment/dislike_icon.html
```diff
@@ -0,0 +1,6 @@
+{% load comment_tags %}
+{% block dislike_img_icon %}
+<svg class="comment-reaction-icon reaction-like" xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill={% if comment|add_one_arg:user|has_reacted:"like" %}"#ffc966"{% else %}"none"{% endif %} stroke="#427297" stroke-width="1.7" stroke-linecap="round" stroke-linejoin="round">
```
Radi85, line 3:
```jinja2
{% if comment|add_one_arg:user|has_reacted:"like" %}
```
We agreed to change the template tags!

Radi85 (changes requested):
Please don't forget to squash your commits.

Radi85:
> Could you guide me with what documentation needs to change and what should be added?. Apart from that, you may check and tell me if anything else needs to be done. In the meantime, I will try to figure out and increase test coverage.

There are not much to do with docs here, I believe, the most important is changing the relaese version to 1.5.0 and point in the docs that the user can react to a comment. However, you could leave this part and I will take care of it.

comment/templates/comment/dislike_icon.html
abhiabhi94: https://github.com/Radi85/Comment/pull/10#discussion_r425442592
I had updated my comment after I saw we couldn't use the simple tag with if tag.

.gitignore
```diff
@@ -53,6 +53,8 @@ docs/_build/
 
 # editor files
 .idea
+.vscode
```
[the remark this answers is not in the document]
abhiabhi94: I also do some front end development. How is PyCharm for it? I was anyway planning to buy a paid version when I have enough money.

abhiabhi94:
> There are not much to do with docs here, I believe, the most important is changing the relaese version to 1.5.0 and point in the docs that the user can react to a comment. However, you could leave this part and I will take care of it.

Okay, should I add the new database file as well?

Radi85:
> > There are not much to do with docs here, I believe, the most important is changing the relaese version to 1.5.0 and point in the docs that the user can react to a comment. However, you could leave this part and I will take care of it.
> 
> Okay, should I add the new database file as well?

If you mean the migration file, yes please.

I got one JS issue when running your code locally.
```js
comment.js:238 Uncaught TypeError: Cannot read property 'style' of undefined
    at fillReaction (comment.js:238)
    at Object.commentReactionDone [as success] (comment.js:281)
```
The issue might be in this line:
```js
if (colorIsNone($likeIcon.style.fill) && colorIsNone($dislikeIcon.style.fill)) {
     $targetReaction.style.fill = fillColor;
}
```

abhiabhi94:
> Please don't forget to squash your commits.

Any particular reason for this? Each individual commit generally has a reason and descriptive message that indicates the reasons for the change.

abhiabhi94:
> If you mean the migration file, yes please.

No the `db.sqlite3` file.

Radi85:
> > If you mean the migration file, yes please.
> 
> No the `db.sqlite3` file.

Doesn't matter since it is used for testing purpose only.

Radi85:
> Any particular reason for this? Each individual commit generally has a reason and descriptive message that indicates the reasons for the change.

I would like to keep git tree clean and I prefer to have one commit per feature.

abhiabhi94:
> I got one JS issue when running your code locally.
> 
> ```js
> comment.js:238 Uncaught TypeError: Cannot read property 'style' of undefined
>     at fillReaction (comment.js:238)
>     at Object.commentReactionDone [as success] (comment.js:281)
> ```
> 
> The issue might be in this line:
> 
> ```js
> if (colorIsNone($likeIcon.style.fill) && colorIsNone($dislikeIcon.style.fill)) {
>      $targetReaction.style.fill = fillColor;
> }
> ```

I have fixed it in the latest commit. It was caused when I was refactoring the templates. I had accidently forgot to change the icon class and filter to the dislike icon file after copying things from the like icon file.

abhiabhi94:
> > Any particular reason for this? Each individual commit generally has a reason and descriptive message that indicates the reasons for the change.
> 
> I would like to keep git tree clean and I prefer to have one commit per feature.

One feature has a lot of sub-features or substantial subtilities that are required for its implementation. I don't think you can put the whole thing in a single commit.

Anyways it is your call. I believe Github gives the owners/maintainers option to squash the commits while merging.

Radi85:
Unexpected behavior here ⬇️ like and dislike both are filled for the same user.
Screenshot 2020-05-16 at 13 45 47

Radi85:
> One feature has a lot of sub-features or substantial subtilities that are required for its implementation. I don't think you can put the whole thing in a single commit.
> 
> Anyways it is your call. I believe Github gives the owners/maintainers option to squash the commits while merging.

Don't worry about it then, I will take care of it.

abhiabhi94:
> Unexpected behavior here arrow_down like and dislike both are filled for the same user.
> Screenshot 2020-05-16 at 13 45 47

Was this behaviour after you clicked on the icons or when the page was rendered?

.gitignore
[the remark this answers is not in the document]
Radi85: PyCharm is pretty cool also for frontend, it has a lot of advantages on code checking for js and css and autocomplete for HTML. However, there are another IDE from Jetbrains called WebStorm and it's specialist with frontend development. I believe the community version is enough and free.

abhiabhi94:
> > > If you mean the migration file, yes please.
> > 
> > 
> > No the `db.sqlite3` file.
> 
> Doesn't matter since it is used for testing purpose only.

Maybe we can put the file in .gitignore to stop it from being tracked, to stop it from causing issues with version control.

Radi85:
> Was this behaviour after you clicked on the icons or when the page was rendered?

After rendreing the page.

.gitignore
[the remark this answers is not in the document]
abhiabhi94: No no, I wanted to pay because I liked the work people put behind it. I like people to pay for such projects so that the opensource projects can support themselves.

Will try PyCharm in the future. Using different IDEs for front end and back end will mess more things inside my head than they are currently are. I am not sure if I can handle that :-p

Radi85:
> Maybe we can put the file in .gitignore to stop it from being tracked, to stop it from causing issues with version control.

I actually keep this db for the demo project so who would like to use app, can directly find an example with less efforts.

abhiabhi94:
> > Was this behaviour after you clicked on the icons or when the page was rendered?
> 
> After rendreing the page.

Are you sure? This means the template filter is not working correctly then. Just to make sure, have you pulled in the latest commits?

abhiabhi94:
I forgot to mention the icons were picked from feathericons.com. Please don't forget to give them credit in the documentation.

Radi85:
> I forgot to mention the icons were picked from https://feathericons.com. Please don't forget to give them credit in the documentation.

Thanks for mentioning. Just small thing regarding the icons, could you please make them with the count text a little bit small?

abhiabhi94:
> > I forgot to mention the icons were picked from https://feathericons.com. Please don't forget to give them credit in the documentation.
> 
> Thanks for mentioning. Just small thing regarding the icons, could you please make them with the count text a little bit small?

Height and width of size 20 for icons look fine to me. Just try it and tell me if it looks good to you.

comment/templates/comment/dislike_icon.html
Radi85: Tested:
```jinja2
{% has_reacted comment user "like" as user_has_reacted %}

{% if user_has_reacted %}"#ffc966"{% else %}"none"{% endif %}
```
I am not happy with filter chain. So far everything looks good!

abhiabhi94: okay, will check and revert to you.

comment/views/reactions.py
abhiabhi94: While I don't agree using a class based view will make the function any cleaner, in this case. I will convert it and commit the change in some time.

Radi85:
> ```
> FAIL: test_create_comment (comment.tests.test_views.CreateCommentTestCase)
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "/home/travis/build/Radi85/Comment/comment/tests/test_views.py", line 45, in test_create_comment
>     self.assertIsNone(response.context.get('comment'))
> AssertionError: reply by test-1: child comment body is not None
> ```

The test is failing due to passing the **reply** comment object to the context data as `comment` which is treated as a parent, I used to differentiate the parent from the child in the context data like so:
```python
context['comment'] = PARENT_COMMENT
context['reply'] = CHILD_COMMENT
```
As I see this is a bad design and I am going to refactor the code.
For now please **comment out** this line `self.assertIsNone(response.context.get('comment'))` in `test_create_comment` **line 45** in order to make the test passing and merge your PR before start refacotring.

Radi85 (approved):

Radi85:
I got `405 (Method Not Allowed)` error when try to react to a comment.

Radi85 (changes requested):
I got 405 (Method Not Allowed) error when try to react to a comment.

abhiabhi94:
> I got 405 (Method Not Allowed) error when try to react to a comment.

Yes, I too get this error now. I had not checked the response by testing it from a browser after the database issue. Since the tests were passing, I assumed it should pass. I will fix this soon and commit.

Radi85 (approved):

Radi85:
Merged in #16

Radi85:
Merged in #16

PR #12, opened 2020-05-16
Change profile app name to resolve name conflics with ipython module

abhiabhi94:
- make other corresponding changes
- fixes #11 

Change some imports order to follow PEP8 conventions.

PR #13, opened 2020-05-16
Abhiabhi94 fix profile app name

Radi85:
Rebase and fix merge conflict of #12


abhiabhi94:
The builds are failing because you haven't to run migrate command.  The migration changes are not reflected in the database. The earlier database had tables under a different name. I forgot to mention this was the reason I had pushed the database file in the last pull request.

tox.ini
```diff
@@ -32,6 +32,7 @@ usedevelop = True
 ignore_outcome =
     djangomaster: True
 commands =
+    python manage.py migrate
```
abhiabhi94, line 35: Shouldn't the flake8 test be placed before anything else?

Radi85: You are right.

PR #18, opened 2020-05-18
Add flag model

### LINKED ISSUE #17: Allow the user reporting a comment
Add flag/report model as suggested in #8

abhiabhi94:
Add flags to comments.

Tasks:
- [x] Add models for Flag and FlagInstance.
    - [x] Add logic for the models.
    - [x] Add signals.
- [x] Make changes in the comments model
    - [x] Add a `post_save` signal that creates a Reaction instance when a comment is created. 
- [x] Add urls and views
    - [x] For django.
    - [x] For REST API.
- [x] Add template tag that
    - [x] Returns if the current user has flagged a particular comment.
    - [x] Returns flag reasons to be rendered for the flag form on the frontend.
- [x]  Fix Frontend
    - [x] Display request response on the frontend.
    - [x] Perform flagging using AJAX request for authenticated users.
    - [x] Redirect unauthenticated users to login page with correct path for the next redirect.
- [x]  Add unit tests for
    - [x] models.
    - [x] views.
    - [x] REST API.
    - [x] template tags.
- [x] Refactor templates directory.
- Fix leftovers
    - [x] Add additional tests to increase coverage.

Fixes #17.


abhiabhi94:
Currently, I expect users to use a setting `COMMENT_FLAGS_ALLOWED` to set the maximum number of flags allowed on a comment. I was also thinkiing of allowing the moderators themselves decide the comments to be flagged if they set this value to 0(although this logic has not been implemented right now).

Questions related to further implementations:
- Do we notify the comment authors if their comment is flagged? If yes, then how?
- Do we notify the moderators when a comment is flagged? If yes, then how?
- Do we allow an option to delete comments permanently, if the number of flags cross the maaximum mark set?

Radi85:
Hey @abhiabhi94,
Sorry for being late, this week was creazy, we had to deliver a product in a week only 😄 and luckily we did it 🚀.

> * Do we notify the comment authors if their comment is flagged? If yes, then how?

No.

> * Do we notify the moderators when a comment is flagged? If yes, then how?

Yes. But I prefer to put this in a new feature. We can use settings attribute to save a list of moderators' emails. Once the comment reaches the allowed flags, an email will be sent to these guys.

> * Do we allow an option to delete comments permanently, if the number of flags cross the maaximum mark set?

For now the moderators can delete the comment from Admin page. This question opens the door to think on implementing user groups and permisions on the comment, we may implement this after finishing the flaggin system.

comment/manager/comments.py
```diff
@@ -1,8 +1,15 @@
+from django.conf import settings
 from django.contrib.contenttypes.models import ContentType
 from django.db import models
 
+ALLOWED_FLAGS = getattr(settings, 'COMMENT_FLAGS_ALLOWED', 10)
 
 class CommentManager(models.Manager):
+    def get_queryset(self):
```
Radi85, line 9: Overriding this method is no longer necessary if we change the field as suggested in the `Flag` model.

comment/models/flags.py
```diff
@@ -0,0 +1,159 @@
...
+        self.count = models.F('count') + 1
+        self.save()
+
+    def decrease_count(self):
+        """Decrease flag count and save the model """
+        self.refresh_from_db()
+        if self.count > 0:
+            self.count = models.F('flag') - 1
+            self.save()
+
```
Radi85, line 53:
```suggestion
@receiver(post_save, sender=Comment)
def create_flag_object(sender, instance, created, raw, using, update_fields, **kwargs):
    if created:
        Flag.objects.create(comment=instance)
```

Radi85 (changes requested):

comment/models/flags.py
```diff
@@ -0,0 +1,159 @@
...
+    State = namedtuple('State', ['value', 'state'])
+
+    # Construct the list of named tuples(can't use list comprehension, scope issues)
+    states = []
+    for state in STATES:
+        states.append(State(*state))
+
+    comment = models.ForeignKey(Comment, on_delete=models.CASCADE, related_name='flags')
+    count = models.PositiveIntegerField(default=0)
+    creator = models.ForeignKey(User, on_delete=models.CASCADE, related_name='flag_owners')
```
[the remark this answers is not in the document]
abhiabhi94: `creator` is the comment author. This was a general name when I thought about the flag model. The flag instance has the name of the user who flagged the content. If you want, I may change the field name to `author`?

comment/models/flags.py
```diff
@@ -0,0 +1,159 @@
...
+    state = models.SmallIntegerField(choices=STATES, default=states[0].value)
+    moderator = models.ForeignKey(User, null=True, on_delete=models.SET_NULL, related_name='flag_moderators')
```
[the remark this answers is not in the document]
abhiabhi94: This was added with the intention to save the user(moderator) who would delete or change the state(a field in the flag model) of the flagged content.

comment/models/flags.py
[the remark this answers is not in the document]
Radi85: Do we really need it as the comment author is already accessible by the `comment` instance. 
```python
comment_author = flag_obj.comment.user
```

comment/models/flags.py
abhiabhi94: I have already added this logic in the comment model. I just forgot to change the name of the function that receives the signal.

comment/models/flags.py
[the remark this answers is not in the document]
abhiabhi94: Yes, in my opinion, it is a utility to the person moderating. But now I think we may add this as a property and not as a model field?

comment/manager/comments.py
abhiabhi94: This method was trying to filter the comments that have been flagged beyond a certain count. How is this concerned with any of such suggested change?

comment/models/flags.py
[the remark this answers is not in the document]
Radi85: Good point. For now the comment cannot be deleted by a user different than the author, We may for now give this permision to the admin until we implement user permissions on the comment. Implementing the permissions gives the developer more flexibility for this case. So I believe we still don't need this field.

comment/models/flags.py
Radi85: Sorry I didn't notice it. I prefer to move this function to the Flad models.

abhiabhi94: We had added this signal in the comments for the reaction model. Hence, I followed the same procedure here.

comment/models/flags.py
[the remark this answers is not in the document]
Radi85: I believe accessing the author from the comment instance doesn't make any confusion. However feel free to use a property for it.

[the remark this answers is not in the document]
abhiabhi94: I was talking in the context of the admin interface.

comment/models/flags.py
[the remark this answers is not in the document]
abhiabhi94: I was talking in context of the admin interface.

comment/manager/comments.py
Radi85: I don't see any reason to override this method and this actually will lead to the following problem:
```python
Comment.objects.all()  # => the result is all comments are FLAGGED
```
If you really think that you need this filter, please use a custom method name instead of overriding the super method.
```suggestion
    def get_flagged_comments(self):
```

comment/models/flags.py
Radi85: I already moved it to models.reactions.

Radi85: I got what you mean. I think the branch you are working on is not up to date 😮
To avoid circular import I moved them. You can make new file signals for such functions in it's worth.

comment/manager/comments.py
abhiabhi94: > ```python
> Comment.objects.all()  # => the result is all comments are FLAGGED
> ```

This function is just filtering out the comments that have a flag count less than a certain number(Currently, the logic is flawed as it returns the reverse i.e. comments that have flag count greater than the number.). I think I might be missing something here as I don't understand your concern.

Radi85: My concern is to NOT overriding the manager method `get_queryset`
Please read this [modifying a manager's initial queryset](https://docs.djangoproject.com/en/3.0/topics/db/managers/#modifying-a-manager-s-initial-queryset)

abhiabhi94: I was overriding this method so that all operations related to comments should only be allowed on instances that have not crossed the allowed flag count criteria. 

These comments should no longer be displayed on a `post`? Am I still missing something?

Radi85: No. I was actually missing something 😃. So you think we should hide the comments that exceed the allowed falgs? why don't we let the developer decide whether hiding them or not??

abhiabhi94: If they want to do it, they can set the count to 0.

comment/models/flags.py
abhiabhi94: My commits were made before you pushed in the respective changes in the other branch. No issues, I will merge and do the needful.

abhiabhi94:
You may review the model. If everything looks good to you, I will start adding the code for views, templates, and other frontend stuff.

Also, a note about `db.sqlite3`. Although it may be helpful for the people using the application, it creates an issue for other contributors to the project. Every time before pushing in the code, I have to do

```bash
git add .
git reset test/db.sqlite3
```
After pulling in the changes, I have to do:

```bash
git checkout test/db.sqlite3 --ours
git add test/db.sqlite3
```
 You don't feel it since you are the owner.

abhiabhi94:
Hey I have written views and templates for this. Before I start writing javascript, I just want to make sure a couple of things.

- Whether a flagged comment is removed if a user clicks on the flag icon or we inform them that their flag has already been recorded and a moderator will look into it?
- If we don't give an option to remove a flag, do we need to colour the flag icon if a comment has been flagged?

abhiabhi94:
I have added views, templatetags, templates, javascript, and CSS for the flagging process. If you could review and give a go-ahead, I will complete the views for API and start writing tests.

Currently the process something similar to:
comment-flags

abhiabhi94:
I was also thinking of moving the template files related to comment, flags, and reactions to their separate directory inside the root for the templates, just to make the directory a bit cleaner. Do you think this would be good?

abhiabhi94:
I have added an element that displays the response of the request and fades away gradually.
This is how it looks now:
comment-flags-1

Radi85:
> * Whether a flagged comment is removed if a user clicks on the flag icon or we inform them that their flag has already been recorded and a moderator will look into it?

For now let's make the flagged commen unremovable.

> * If we don't give an option to remove a flag, do we need to colour the flag icon if a comment has been flagged?

This would be great to give the user hint that he flagged the comment. Other users should not see the flagged comment.

Radi85:
> I have added views, templatetags, templates, javascript, and CSS for the flagging process. If you could review and give a go-ahead, I will complete the views for API and start writing tests.

Hey @abhiabhi94,

Well done so far. I will try to review today

Radi85:
> I was also thinking of moving the template files related to comment, flags, and reactions to their separate directory inside the root for the templates, just to make the directory a bit cleaner. Do you think this would be good?

Great idea. Go ahead please.

Radi85:
> I have added an element that displays the response of the request and fades away gradually.

It looks beautiful and this is much better that a JS alert. One thing only, moving the message below the Comment title may be better.

comment/models/flags.py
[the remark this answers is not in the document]
Radi85: TBH I don't see any reason for the moderator field here. The flagged comments will be reported to the list of moderators in settings or for now to the admin. Later we can add the moderators to a group which can delete the flagged comments only. I can implement this after merging your PR but for now please remove this field.

Radi85 (changes requested):
Other than the moderator field, everything looks good

abhiabhi94:
I have moved the response element just above the comment title now. It looks something similar to:
comment-flags-3

comment/templates/comment/child_comment.html
```diff
@@ -1,5 +1,5 @@
 
-<blockquote class="js-child-comment blockquote text-wrap mb-4">
+<div class="js-child-comment blockquote text-wrap mb-4">
     <hr/>
     {% include 'comment/image.html' with comment=comment %}
-</blockquote>
```
abhiabhi94, line 5: As far as I can see the [documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/blockquote), I believe the blockquote element was misplaced here.

abhiabhi94:
I had added functions to provide backward compatibility for previous comments i.e. allow reactions and flagging for comments that were added before this update.

The API tests seem more like integration tests to me and would ideally require some refactoring. May be some other day.

comment/api/views.py
```diff
@@ -64,6 +65,7 @@ def post(self, request, *args, **kwargs):
         if not reaction_type:
             data = {'error': 'This is an invalid reaction type'}
             return Response(data, status=status.HTTP_400_BAD_REQUEST)
+        reaction = SetReaction.get_reaction_object(comment)
```
Radi85, line 68: Why was this line added? I don't see any further use of `reaction` object

comment/api/serializers.py
```diff
@@ -89,6 +89,15 @@ def get_dislikes(obj):
         if hasattr(obj, 'dislikes'):
             return obj.dislikes
 
+    @staticmethod
+    def get_is_flagged(obj):
+        if hasattr(obj, 'flag'):
```
Radi85, line 94: The object already has `is_flagged` property just return this property instead of repeating the same method.
```python
return obj.is_flagged
```

Radi85 (changes requested):

comment/templates/comment/child_comment.html
Radi85: This was used to get the current comment style, I believe I got this block from bootstrap. However, feel free to update the style and the templates.

Radi85:
> I have added functions to provide backward compatibility for previous comments i.e. allow reactions and flagging for comments that were added before this update.

I saw them, well done. Thanks!

> The API tests seem more like integration tests to me and would ideally require some refactoring. I was feeling too lazy right now to fix them, maybe some other day.

No worries, I will cover all the missing tests after merging.

comment/api/views.py
abhiabhi94: This is for providing reaction objects for previous comments. Earlier I had used `getorcreate` function but you transformed this in the latest release. To provide backward compatibility, this is necessary.

comment/api/serializers.py
abhiabhi94: Previous comments will not have a `flag` object.

abhiabhi94:
> > The API tests seem more like integration tests to me and would ideally require some refactoring. I was feeling too lazy right now to fix them, maybe some other day.
> 
> No worries, I will cover all the missing tests after merging.

I was talking about the tests written earlier for comments.

comment/api/views.py
Radi85: I undertood what you intend to do. The code here is a little bit confusing can you please adjust it as follows:
```suggestion
        reaction_obj = SetReaction.get_reaction_object(comment)
        ReactionInstance.objects.set_reaction(
            user=request.user,
            reaction=reaction_obj,
            reaction_type=reaction_type
        )
```

comment/views/flags.py
```diff
@@ -0,0 +1,49 @@
...
+from django.views.decorators.http import require_POST
+
+from comment.models import Comment, Flag, FlagInstance
+
+
+@method_decorator(require_POST, name='dispatch')
+class SetFlag(LoginRequiredMixin, View):
+
+    @staticmethod
+    def get_flag_object(comment):
```
Radi85, line 17: Can you please move this method to Flag manager? I believe this logic is in the wrong place.
Getting an object should be from the model class not the view.

comment/views/reactions.py
```diff
@@ -1,17 +1,27 @@
...
 
-from comment.models import Comment, ReactionInstance
+from comment.models import Comment, Reaction, ReactionInstance
 
 
 @method_decorator(require_POST, name='dispatch')
 class SetReaction(LoginRequiredMixin, View):
 
+    @staticmethod
+    def get_reaction_object(comment):
```
Radi85, line 17: Also applied here as for the flag.

comment/api/serializers.py
Radi85: I am still seeing the old code :) I will resolve it once done.

Radi85 (changes requested):

comment/views/flags.py
abhiabhi94: I had though of placing this inside the Comment manager or model but that would cause circular imports. Would it be best placing this inside the flag model, design-wise?

Radi85: Sure. The best place for a method that returns a model object is the model manager.
```django
Flag.objects.get_for_comment(comment=comment)
Reaction.objects.get_for_comment(comment=comment)
```

abhiabhi94: I have made the change. In an ideal world, you would want to put the logic inside Comment manager because these objects are related to the instance of the comment model but that creates issues of circular import.

Radi85: Why would the circular import happen? To get the model class inside the manager you can use `self.model` instead of importing it.

abhiabhi94: We need to use the Reaction or Flag class inside the Comment Manager. It would create circular imports as Comment is being used in Reaction and now we can't use Reaction inside Comment.

Radi85 (approved):

comment/managers/flags.py
```diff
@@ -0,0 +1,105 @@
...
+        if action == 'delete':
+            instance = self.get(flag=flag, user=user)
+            instance.delete()
+            created = False
+        else:
+            self._clean(reason, info)
+            self.create(flag=flag, user=user, reason=reason, info=info)
+            created = True
+
+        return created
```
Radi85, line 105: hey @abhiabhi94,
I have found an issue with the current design for flagging. The issue is derived from `action` variable when the user intends to flag a comment. IMHO we don't need this variable explicitly at all because the action is already known.
* If the user has not flagged a comment => the action will be `create`
* else `delete`
we can just implement a simple check to find out the action type instaed of passing a variable within JS and html and this will solve the api issue as well.
```suggestion
        reason = kwargs.get('reason', None)
        info = kwargs.get('info', None)
        instance, created = self.get_or_create(user=user, flag=flag)
        if not created:
            instance.delete()
        else:
            self._clean(reason, info)
            instance.reason = reason
            instance.info = info
            instance.save()

        return created
```

Radi85 (changes requested):

comment/managers/flags.py
abhiabhi94: I had used this to stop the abuse of the API. Now that I think of it, I don't think we can stop abuse just by doing this. Your suggestion seems legit to me. I will make the changes and remove the unnecessary variable. I will make the changes and push in a while.

abhiabhi94: After making the changes, I now remember why I added the action argument. So in this case, for handling the wrong values of reason or info, we would have to catch the exception and delete the instance. This didn't seem like a nice way to deal with such cases to me. If you still want the change, I will push it.

abhiabhi94:
There were a couple of bugs in the current code.

- **Overriding the save function**: In the current state, it causes an increase in the count even when one of the model fields is updated. I have moved this logic to a `post_save` signal.

- **Sometimes the text field is saved inside the browser**: A user firstly selects something else and enters some info(let it be `'Hi'`). They then remove the flag on the comment. They again flag the comment and choose a different reason for flagging. In this case, the text field content is saved as the one before( `'Hi'`) in the text field. This gets saved into the database because the code doesn't check for the `info` value in case a reason other than something else is chosen. Changing some logic in the clean function for the `Flag Manager` fixes this.

comment/managers/flags.py
Radi85: > So in this case, for handling the wrong values of reason or info, we would have to catch the exception and delete the instance.

I am not sure that I understood what you mean exactly, but what I see is that the action has nothing to do with reason and info here. As I said in case you need this variable, just keep it within django code and don't pass it from the JS or the template.
```python
instance, created = self.get_or_create(user=user, flag=flag)
if created:
    action = 'create'
```
I really don't see anymore reason for using the action variable since `created` is available for us.

There is an issue with the api when a user try to flag a comment, I will fix this after removing the action variable.

abhiabhi94: Okay. Suppose a user tinkers something with the frontend and pass a wrong reason to the API. So according to your suggestion, `get_or_create` will create an instance and the execution will be passed to the else branch. Now the clean method throws a Validation error for the wrong value of reason. If put a `try except` block to catch this error, I will have to delete the instance now that this instance is no longer useful.

I didn't think this was a great method to deal with wrong values.

abhiabhi94: > There is an issue with the api when a user try to flag a comment, I will fix this after removing the action variable.

What is the issue?

Radi85: > Okay. Suppose a user tinkers something with the frontend and passes a wrong reason to the API. So according to your suggestion, `get_or_create` will create an instance and the execution will be passed to the else branch. 

Simply check a valid reason before calling `get_or_create`.

Radi85: > What is the issue?

* How the data (reason, action and info) are passed when using api? You are using `**request.POST.dict()` but actually these data are not available in this dict, therefore, when you try to post to `comments/<int:pk>/flag/` you will always get an error, `None is not a valid action`
* Tried to pass these data as qurey string it worked somehow, but when posting twice with the same action, an `IntegrityError` is raised, therefore, I requested to remove action variable. IMHO it's not a good design.

Radi85: > * How the data (reason, action and info) are passed when using api? You are using `**request.POST.dict()` but actually these data are not available in this dict, therefore, when you try to post to `comments/<int:pk>/flag/` you will always get an error, `None is not a valid action`

Well, I just checked your tests and the data are passed correctly. The issue is that I used the GUI api of the restframework. This can be ignored now.

abhiabhi94: > > Okay. Suppose a user tinkers something with the frontend and passes a wrong reason to the API. So according to your suggestion, `get_or_create` will create an instance and the execution will be passed to the else branch.
> 
> Simply check a valid reason before calling `get_or_create`.

For delete, no reason will be passed. This will create new issues.

abhiabhi94: > > What is the issue?
> 
>     * How the data (reason, action and info) are passed when using api? You are using `**request.POST.dict()` but actually these data are not available in this dict, therefore, when you try to post to `comments/<int:pk>/flag/` you will always get an error, `None is not a valid action`
> 
>     * Tried to pass these data as qurey string it worked somehow, but when posting twice with the same action, an `IntegrityError` is raised, therefore, I requested to remove action variable. IMHO it's not a good design.

Do you get why the variable was used here?  Also, the parameters need to be passed via a query string.

Radi85: Delete the instance if no reason passed:
```python
        reason = kwargs.get('reason', None)
        info = kwargs.get('info', None)
        instance, created = self.get_or_create(user=user, flag=flag)
        if not created or not reason:
            instance.delete()
        else:
            self._clean(reason, info)
            instance.reason = reason
            instance.info = info
            instance.save()
        return created
```

abhiabhi94: How do you deal with a wrong value of reason? You need to delete it in this case too.

abhiabhi94: Also, you don't raise an error in the case when no reason is passed to flag a comment. Since you don't have an action variable, the code doesn't really know when to raise an exception.

Radi85: Ok then:
```python
        reason = kwargs.get('reason', None)
        info = kwargs.get('info', None)
        if reason:
             self._clean(reason, info)  # an exception will be raised if the reason not valid
        # here means that we have a valid reason or no reason passed
        instance, created = self.get_or_create(user=user, flag=flag)
        if not created or not reason:
            instance.delete()
        else:
            instance.reason = reason
            instance.info = info
            instance.save()
        return created
```

abhiabhi94: Even in this case, the clean function will not run when no reason is passed to flag a comment. You want it to run everytime when flagging a comment, and let clean handle the exception handling.

Radi85: Why do I need to run clean if reason not passed. It must be abvoius that when there is no reason passed, the action must be delete if the instance exists.

Radi85: We have 2 cases only:
* reason is passed, the action is create if the reason valid, and.
* reason is not passed, the action is delete if the instance exists.

abhiabhi94: Do we assume that there are no people in this universe who would tinker with the API? If someone tinkers with the API from the browser, an exception should be raised in my opinion if not all parameters are satisfied.

Radi85: I am not sure how this is related to our subject? However, you may leave it this way and I will refactore it before merging.

abhiabhi94: Okay, I will repeat the example I gave above. 

If someone tinkers with the API and sends a request to the server without a reason should we just delete the instance or raise an exception if the parameters are not according to our need?

In this case, we delete the instance, because we have no way to know if the request for creation or deletion.

Radi85: > Okay, I will repeat the example I gave above.
> 
> If someone tinkers with the API and sends a request to the server without a reason should we just delete the instance or raise an exception if the parameters are not according to our need?

We will delete the instance, same user cannot flag a comment twice with a reason so this is abvoius that a request without a reason for particular user means deleting the flag.

> In this case, we delete the instance, because we have no way to know if the request was for creation or deletion.

We already know the type of the request. A flagged comment cannot be FLAGGED again by the same user, therefore, any further request from the same user means delete the existing flag.

abhiabhi94:
Okay, so you are not considering the proposition that the someone might try
to tinker with the API and send a request without the reason parameter.

In this case, we just delete the instance. In my opinion it seems better to
raise an exception and return an appropriate response without touching the
database in this case.

Radi85:
> Okay, so you are not considering the proposition that the someone might try to tinker with the API and send a request without the reason parameter. In this case, we just delete the instance. In my opinion it seems better to raise an exception and return an appropriate response without touching the database in this case.

Someone? Only authenticated user can flag a comment and only the author of the flag instance can delete the flag!! I am not sure why we struggle here? Someone cannot delete an arbitrary flag instance.

```python
# the instance which might be deleted later is realted to particular authenticated USER not an Anonymous.
instance, created = self.get_or_create(user=user, flag=flag)
```

Radi85 (approved):

Radi85:
Thanks a lot @abhiabhi94 for your efforts.
Merged in #22

PR #35, opened 2020-06-24
update flag states

Radi85:
- The comment admin and moderator can now toggle between the flag states of a flagged comment.
- Add function to check and update the flag state after each migrate
command.
- Extend the API to cover all GUI actions


comment/managers/comments.py
```diff
@@ -12,8 +12,7 @@ def all_exclude_flagged(self):
         if not allowed_flags or show_flagged:
             return super().get_queryset()
 
-        return super().get_queryset().select_related('flag').annotate(
-            flags=models.F('flag__count')).filter(flags__lte=allowed_flags)
+        return super().get_queryset().exclude(flag__state__exact=2)
```
abhiabhi94, line 15: Instead of using raw numbers maybe we may use Enum choices.

comment/models/flags.py
```diff
@@ -12,12 +15,23 @@
...
 class Flag(models.Model):
+    UNFLAGGED = 1
+    FLAGGED = 2
+    REJECTED = 3
+    RESOLVED = 4
+    STATES_CHOICES = [
+        (UNFLAGGED, 'Unflagged'),
+        (FLAGGED, 'Flagged'),
+        (REJECTED, 'Flag rejected by the moderator'),
+        (RESOLVED, 'Comment modified by the author'),
```
abhiabhi94, line 26: Earlier gettext_lazy was used with the raw message for the state. It would help users from non-english language to easily port this in their language.

```python
from django.utils.translation import gettext_lazy as _

STATE_CHOICES = [
        (UNFLAGGED, _('Unflagged')),
        (FLAGGED, _('Flagged')),
        (REJECTED, _('Flag rejected by the moderator')),
        (RESOLVED, _('Comment modified by the author')),
]

File: comment/models/comments.py [hunk 3]
```diff
@@ -58,8 +57,25 @@ def dislikes(self):
...
     def is_flagged(self):
         if hasattr(self, 'flag'):
-            self.flag.refresh_from_db()
-            allowed_flags = getattr(settings, 'COMMENT_FLAGS_ALLOWED', 0)
-            if allowed_flags:
-                return self.flag.count > allowed_flags
+            if not self.flag.is_flag_enabled:
+                return False
+            return self.flag.state != self.flag.UNFLAGGED
+        return False
```
abhiabhi94, line 63: This can be changed to:

```python

if hasattr(self, 'flag'):
    if self.flag.is_flag_enabled:
        return self.flag.state != self.flag.UNFLAGGED
return False

abhiabhi94:
Also, I had used 1 one single setting variable for allowing flags and their count. I think you have moved them to separate variables.

We can ask the developer to set the value as 0 or False(whatever you feel like) to check if flags are allowed. This will help in moving the changes to one variable.

comment/models/flags.py
Radi85: You are right. I will open an issue for translating to cover everything.

comment/managers/comments.py
Radi85: Yes I will look at this in future code refactoring, I tried to get `Flag.Flagged` but got a circular import.

Radi85:
> Also, I had used 1 one single setting variable for allowing flags and their count. I think you have moved them to separate variables.
> 
> We can ask the developer to set the value as 0 or False(whatever you feel like) to check if flags are allowed. This will help in moving the changes to one variable.

The new variable is for new purpose.
So we have one for allowed flag counts and if it's 0 then the flagging system will be disabled
the second one is to keep the flagged comment visible to all users. We previously hided the flagged comments

comment/models/comments.py
Radi85: Your suggestion is not valid here because the developer may disable flagging and `is_flagged` will still be true and the comments will appear as flagged even if the feature is disabled.

abhiabhi94: I'm not sure I understand your argument since I've not seen all the changes in depth. What I suggested was a simple refactoring of your code, which does the exact same thing as your code or does it?

comment/managers/comments.py
abhiabhi94: Yes, that is the reason I had used the statues in managers.

comment/models/comments.py
Radi85: Simply: `self.flag.state != self.flag.UNFLAGGED` can be False or True, even if the flagging system is disabled.
I want `is_flagged` to be always false when the flagging system is disabled.

abhiabhi94: Okay, now I understand. Ihas mistakenly added the `hasattr` check in my comment. I have [edited](https://github.com/Radi85/Comment/pull/35#discussion_r444945408) it now. Everything else is fine.

abhiabhi94:
Okay. Suppose if I set allowed flag count to 10, I obviously want the comment below this number to be hidden. Do I miss something here?

One use case that I can think of is notifying the moderators after this count for a review. I don't think we have implemented this till now.

Radi85:
> Okay. Suppose if I set allowed flag count to 10, I obviously want the comment below this number to be hidden. Do I miss something here?

The comment that has flags count above this number will be hidden and here comes the new variable, maybe the developer doesn't want that comment to be hidden then he can simply set `COMMENT_SHOW_FLAGGED=True`
I believe everything will be clear when you make a quick tour around the code. [The docs](https://github.com/Radi85/Comment#3-enable-flagging)

> One use case that I can think of is notifying the moderators after this count for a review (before which the comments should/shouldn't be visible). I don't think we have implemented this till now.

We haven't.

abhiabhi94:
Okay, One of them is used to enable the flagging system.

Also, In my opinion, the migration of comments should ideally be handled via a command to `manage.py`. I was feeling too lazy while adding this feature. I will work on this sometime in the future.

PR #38, opened 2020-07-01
Add permalink to comment

### LINKED ISSUE #36: Add permalink to each comment
_Originally posted by @abhiabhi94 in https://github.com/Radi85/Comment/issues/33#issuecomment-648842742_

abhiabhi94:
- add a `urlhash` to the model
- write migrations to make the changes compatible for previous comments
- add template to render permalink icon alongside comment

Change `render_content` template tag
- complete comment object is passed now so that urlhash can be obtainer easily.

Add tests
- all a base class for testing custom migration code.
- Add unit tests for general purpose `utils`.

TODO:

- [ ]  Beautify frontend so that the comment may be highlighted when it's link is accessed. 
- [x]  Write remaining tests (especially for migrations).

Current scenario:

comment-permalink

Fixes #36


abhiabhi94:
Hey Radico.

Another issue with providing the permalink icon only on hover is that it will restrict its access for touch devices.

comment/templates/comment/comments/content.html
```diff
@@ -1,9 +1,12 @@
...
     {% if text_2 %}
         {{ text_1 }}
         <span class="d-none">{{ text_2 }}</span>
         <button class="js-read-more-btn btn btn-link btn-xs" style="font-size: 14px">read more...</button>
     {% else %}
         {{ text_1 }}
     {% endif %}
+    {% block comment_urlhash %}
+    {% include "comment/comments/urlhash.html" with urlhash=urlhash %}
+    {% endblock comment_urlhash %}
```
Radi85, line 11: I would prefer to add the permalink icon to the beginning of the comment with slightly reducing its size.
```suggestion
    {% block comment_urlhash %}
    {% include "comment/comments/urlhash.html" with urlhash=urlhash %}
    {% endblock comment_urlhash %}
    {% if text_2 %}
        {{ text_1 }}
        <span class="d-none">{{ text_2 }}</span>
        <button class="js-read-more-btn btn btn-link btn-xs" style="font-size: 14px">read more...</button>
    {% else %}
        {{ text_1 }}
    {% endif %}
```

comment/migrations/0008_comment_urlhash.py
```diff
@@ -0,0 +1,49 @@
...
+            instance.urlhash = generate_urlhash()
+
+
+def set_default_urlhash(apps, schema_editor):
+    Comment = apps.get_model('comment', 'Comment')
+    comment_obj = Comment()
+    for comment in Comment.objects.all():
+        set_unique_urlhash(Comment, comment_obj, comment)
+        comment.save(update_fields=['urlhash'])
+
```
Radi85, line 29: For compatibility with sonarlint, please user ``comment_model`` variable instead of ``Comment``
```suggestion
def set_unique_urlhash(comment_model, obj, instance):
    if not instance.urlhash:
        hash_val = generate_urlhash()
        instance.urlhash = hash_val
        while comment_model.objects.filter(urlhash=hash_val).exists():
            instance.urlhash = generate_urlhash()


def set_default_urlhash(apps, schema_editor):
    comment_model = apps.get_model('comment', 'Comment')
    comment_obj = comment_model()
    for comment in comment_model.objects.all():
        set_unique_urlhash(comment_model, comment_obj, comment)
        comment.save(update_fields=['urlhash'])

```

Radi85 (changes requested):

comment/models/comments.py
```diff
@@ -17,6 +17,7 @@ class Comment(models.Model):
     edited = models.DateTimeField(auto_now=True)
 
     objects = CommentManager()
+    urlhash = models.CharField(max_length=objects.len_urlhash, unique=True, editable=False)
```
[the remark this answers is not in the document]
abhiabhi94: `objects` won't be defined here then  `urlhash = models.CharField(max_length=objects.len_urlhash, unique=True, editable=False)`

comment/templates/comment/comments/content.html
abhiabhi94: That would make the icon appear to the left. It would be different from what I've seen at different places.

comment/migrations/0008_comment_urlhash.py
abhiabhi94: I will change the variable name but since I currently don't use `SonarLint` I really don't understand the exact issue here.

comment/models/comments.py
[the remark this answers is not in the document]
Radi85: I don't find any reason for using dynamic value of max_length. So please adjust it or move `len_urlhash ` to the model directly. I would love to keep the manager for its original purpose which is defining the model interface with DB.

comment/migrations/0008_comment_urlhash.py
Radi85: Me neither but sonarLint complains 😄 

Screenshot 2020-07-02 at 13 32 02

comment/models/comments.py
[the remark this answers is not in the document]
abhiabhi94: If the user want they can change the suffix and prefix to the `urlhash` according to their choice. Any changes there will be reflected in the model and generate a new migration file.

It would otherwise create an issue if the values are changed.

comment/migrations/0008_comment_urlhash.py
abhiabhi94: This is from the official django documentation. It can be found under the heading https://docs.djangoproject.com/en/3.0/howto/writing-migrations/#migrations-that-add-unique-fields
```python
# Generated by Django A.B on YYYY-MM-DD HH:MM
from django.db import migrations
import uuid

def gen_uuid(apps, schema_editor):
    MyModel = apps.get_model('myapp', 'MyModel')
    for row in MyModel.objects.all():
        row.uuid = uuid.uuid4()
        row.save(update_fields=['uuid'])

class Migration(migrations.Migration):

    dependencies = [
        ('myapp', '0004_add_uuid_field'),
    ]

    operations = [
        # omit reverse_code=... if you don't want the migration to be reversible.
        migrations.RunPython(gen_uuid, reverse_code=migrations.RunPython.noop),
    ]
```

comment/models/comments.py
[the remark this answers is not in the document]
Radi85: I am not sure if I understood your idea but I believe it's not related to our issue here since I am just requsting to move a variable to another file. You can for example use `max_length=50` and restict the developer to not use more than this if you think that the length might exceeded or as I said move this valiable to utils and use it in managers and models.

comment/migrations/0008_comment_urlhash.py
abhiabhi94: From what I see in the image, the rule is for normal variables and not Classes as such.

comment/templates/comment/comments/content.html
Radi85: Yes I know. Let's be special case 😄

comment/models/comments.py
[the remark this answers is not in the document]
abhiabhi94: Okay, if you see the setting file there are variables `COMMENT_URL_PREFIX` and `COMMENT_URL_SUFFIX` which are used to make the urlhash. The length is calculated dynamically accordingly.

But as you said, we may set the `max length=50` and ask the developer to not use suffix, prefix and id length that when combined exceed this length, somewhere in the documentation. Are you okay with it?

comment/templates/comment/comments/content.html
Radi85: I tried it and looks cool

Screenshot 2020-07-02 at 13 45 35

comment/migrations/0008_comment_urlhash.py
Radi85: Yes please. Simply django code recommendation is not compatible with sonarLint

comment/templates/comment/comments/content.html
abhiabhi94: It just looks different and a bit confusing to me. I'm short of better words right now :)

Also, you probably missed this part in my comment above
> Another issue with providing the permalink icon only on hover is that it will restrict its access for touch devices.

comment/migrations/0008_comment_urlhash.py
abhiabhi94: Okay, if you consider SonarLint's recommendation to be above Django's, I will change it.

comment/templates/comment/comments/content.html
Radi85: We should not implement the hovering feature

abhiabhi94: What size and padding have you used for the icon?

Radi85: I used your icon without changing any styles.

comment/models/comments.py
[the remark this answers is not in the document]
Radi85: Yup.

comment/tests/test_migrations.py
```diff
@@ -37,4 +39,21 @@ def test_adjust_flagged_comments(self):
...
+    migrate_from = '0007_auto_20200620_1259'
+    migrate_to = '0008_comment_urlhash'
+
+    def setUpBeforeMigration(self, apps):
+        comment_model = apps.get_model('comment', 'Comment')
+        self.comment_id = comment_model.objects.create(
+            content_type_id=self.ct_object.id,
+            object_id=1,
+            content='test urlhash',
+            user_id=self.user.id
```
abhiabhi94, line 52: This should ideally be
```python

self.comment_id = comment_model.objects.create(
    content_object=self.ct_object,
    content='hi',
    user=self.user).id
```
But this was causing weird issues, probably because the model class here is `fake`. If you figure out and solve the issue, please inform me as well.

comment/tests/test_utils.py
```diff
@@ -81,3 +82,38 @@ def test_get_comment_context_data(self):
...
+    def test_id_generator_length(self):
+        self.assertEqual(self.len_id, len(id_generator()))
+
+    def test_id_generator_generates_different_ids(self):
+        self.assertNotEqual(id_generator(), id_generator())
+
+    def test_id_generator_prefix(self):
+        prefix = 'comment'
+        output = id_generator(prefix=prefix)
+        self.assertEqual(True, output.startswith(prefix))
```
abhiabhi94, line 102: I see that you have removed most `self.assertEqual(True, condition)` with `self.assertTrue(condition)`. Both do the same work, in general `self.assertEqual` gives more verbose message on failing.

comment/models/comments.py
```diff
@@ -13,6 +14,11 @@ class Comment(models.Model):
     object_id = models.PositiveIntegerField()
     content_object = GenericForeignKey('content_type', 'object_id')
     content = models.TextField()
+    urlhash = models.CharField(
+        max_length=len(settings.COMMENT_URL_PREFIX + settings.COMMENT_URL_SUFFIX) + settings.COMMENT_URL_ID_LENGTH,
```
Radi85, line 18: We agreed to use `max_length=50`.
I just realized that there will be significant issue with this design, whenever the developer change the `COMMENT_URL_*` settings, the `max_lenght` will be changed and the model will be updated, new migration file will be ceated and this will mess the migration history up.

Radi85 (changes requested):
Well done @abhiabhi94 ,
Many thanks for adding base class for migrations tests.
I still have one concren on the `urlhash` field

Radi85:
I see that there are few lines need to be tested? Can we cover them?

Screenshot 2020-07-04 at 11 38 42

comment/tests/test_utils.py
Radi85: you can use `assertEqual` for almost any case, but I would like to be specific when asserting booleans with `assertTrue` and `assertFalse`. I don't think that we need the verbose message here since we expect 2 values only `True` or `False`.

comment/tests/test_migrations.py
Radi85: I will look at it.

comment/models/comments.py
abhiabhi94: The migration history won't be messed up as long as these changes are made after running our migrations. I have used static values in our migrations.

comment/tests/test_utils.py
abhiabhi94: It is more helpful when debugging in case the test fails.

abhiabhi94:
> I see that there are few lines need to be tested? Can we cover them?
> 
> Screenshot 2020-07-04 at 11 38 42

The lines that fail are set

> I see that there are few lines need to be tested? Can we cover them?
> 
> Screenshot 2020-07-04 at 11 38 42
These were the lines that were left:
```python
while model.objects.filter(urlhash=hash_val).exists():
            instance.urlhash = generate_urlhash()
```

I think I can try to test them under models by setting this property `urlhash` and then calling the save method. I'm not really sure how I can do this for the migration case.

I will test for the models in a bit and revert.

abhiabhi94:
> > I see that there are few lines need to be tested? Can we cover them?
> > Screenshot 2020-07-04 at 11 38 42
> 
> These were the lines that were left:
> 
> ```python
> while model.objects.filter(urlhash=hash_val).exists():
>             instance.urlhash = generate_urlhash()
> ```
> 
> I think I can try to test them under models by setting this property `urlhash` and then calling the save method. I'm not really sure how I can do this for the migration case.
> 
> I will test for the models in a bit and revert.

I did try this:
```python
if not self.urlhash or (self.__class__.objects.filter(urlhash=self.urlhash).exists()):
            hash_val = self.__class__.objects.generate_urlhash()
            self.urlhash = hash_val
            while self.__class__.objects.filter(urlhash=hash_val).exists():
                self.urlhash = self.__class__.objects.generate_urlhash()
```
I can obviously test this code for uniqueness of hash by setting a hash value of one comment to that of another and calling the `save` method for the first. This still doesn't really solve the problem as the last branch under the while loop is still not covered. 

For it to be assessed the function `id_generator` function needs to give the same values twice. That function generates a random string sequence of lowercase English characters.

I'm not sure how we can test it. Any suggestions from you would be great.

Radi85:
> ```python
> while self.__class__.objects.filter(urlhash=hash_val).exists():
> ```
> I'm not sure how we can test it. Any suggestions from you would be great.

I will have a deeper look in an hour, but what I can think about is mocking this value `self.__class__.objects.filter(urlhash=hash_val).exists()`

Radi85:
> For it to be assessed the function `id_generator` function needs to give the same values twice. That function generates a random string sequence of lowercase English characters.

Hmm, instead of mocking this value `self.__class__.objects.filter(urlhash=hash_val).exists()` as I said in the previous comment, you can mock `id_generator` directly to give you same value of existing one.

comment/models/comments.py
Radi85: It does really. I just changed the settings `COMMENT_URL_ID_LENGTH = 20` and run migrations and new migrations file was created here:
```python
# Generated by Django 3.0.7 on 2020-07-04 11:37

from django.db import migrations, models


class Migration(migrations.Migration):

    dependencies = [
        ('comment', '0008_comment_urlhash'),
    ]

    operations = [
        migrations.AlterField(
            model_name='comment',
            name='urlhash',
            field=models.CharField(editable=False, max_length=28, unique=True),
        ),
    ]

```

comment/tests/test_migrations.py
Radi85: I believe as you said because of using fake model. I cannot find another explaination.

comment/migrations/0008_comment_urlhash.py
```diff
@@ -0,0 +1,48 @@
+from django.db import migrations, models
+
+from comment.conf import settings
```
Radi85, line 3: Please remove unused import

comment/migrations/0008_comment_urlhash.py
```diff
@@ -0,0 +1,48 @@
...
+from comment.utils import id_generator
+
```
Radi85, line 5: Add empty line for pep8 compatibility.

comment/migrations/0008_comment_urlhash.py
```diff
@@ -0,0 +1,48 @@
...
+def generate_urlhash():
+    return id_generator(
+        prefix='comment',
+        len_id=8,
+        suffix=''
+        )
+
+
+def set_unique_urlhash(model, obj, instance):
```
Radi85, line 14: `obj` is unused param, please remove it.

Radi85 (changes requested):

comment/models/comments.py
abhiabhi94: This doesn't mess up with the migration history. The file is generated since the constraints in the database have to be changed. The history is intact.

abhiabhi94:
> > For it to be assessed the function `id_generator` function needs to give the same values twice. That function generates a random string sequence of lowercase English characters.
> 
> Hmm, instead of mocking this value `self.__class__.objects.filter(urlhash=hash_val).exists()` as I said in the previous comment, you can mock `id_generator` directly to give you same value of existing one.

I will test this in a bit and revert.

comment/tests/test_migrations.py
abhiabhi94: Also I have used `TransactionTestCase` instead of `TestCase` since it was causing issues with atomic requests.

abhiabhi94:
> > > For it to be assessed the function `id_generator` function needs to give the same values twice. That function generates a random string sequence of lowercase English characters.
> > 
> > 
> > Hmm, instead of mocking this value `self.__class__.objects.filter(urlhash=hash_val).exists()` as I said in the previous comment, you can mock `id_generator` directly to give you same value of existing one.
> 
> I will test this in a bit and revert.

Hey I did test this out. While I can mock the output of the function `comment.managers.CommentManager.generate_urlhash`, the control will obviously get struck in an infinite loop since the `urlhash` isn't unique. How do I get around it?

Radi85:
> Hey I did test this out. While I can mock the output of the function `comment.managers.CommentManager.generate_urlhash`, the control will obviously get struck in an infinite loop since the `urlhash` isn't unique. How do I get around it?

I suppose you are using the returned value as follow:
```python
mocked_func.return_value = YOUR_VALUE
```
In this case the returned value will be always same, instead use `side_effect` like so:
```python
mocked_func.side_effect = [first_value, second_value]  # for example [True, False]
```
in this case the while loop will break.

comment/models/comments.py
Radi85: I am not sure why we are arguing on this?! I don't want a migration file to be created whenever the developer change his settings, this is absolutely not a good design.

Radi85 (changes requested):

comment/models/comments.py
Radi85: BTW, the developers who use our app should not be able to create or change any of the migration files.

comment/tests/test_models.py
```diff
@@ -110,6 +112,24 @@ def test_has_resolved_state(self, mocked_hasattr):
...
+            side_effect=[True, False]
+        ):
+            comment_1.save()
+
+            self.assertEqual(comment_1.urlhash, True)
+
+        comment_1.urlhash = None
+        comment_1.save()
+        comment_1.refresh_from_db()
+        self.assertNotEqual(comment_1.urlhash, comment_2.urlhash)
```
abhiabhi94, line 131: This is just a workaround to push the changes. I tried different sort of stuff to fix this but was unable to do it till now. I have give this close to 3 hours and am certainly out of patience now. It would be nice if you could help and fix this.

comment/models/comments.py
abhiabhi94: > I am not sure why we are arguing on this?! I don't want a migration file to be created whenever the developer change his settings, this is absolutely not a good design.

Okay. 50 is a reasonable length here.

Also, It would be better if we use gender-neutral pronouns like they instead of his/her.

abhiabhi94: Maybe we also need to put a test somewhere that checks this constraint and asks the developer to adhere if set values that go beyond 50. I was not sure where to put this in a file that is triggered when the settings are changed.

abhiabhi94: > BTW, the developers who use our app should not be able to create or change any of the migration files.

As long as the code is open-source they obviously can :-p Although, I get your point in this context.

comment/models/comments.py
```diff
@@ -38,6 +43,17 @@ def replies(self, include_flagged=False):
             return self.__class__.objects.filter(parent=self).order_by('posted')
         return self.__class__.objects.all_exclude_flagged().filter(parent=self).order_by('posted')
 
+    def _set_unique_urlhash(self):
+        if not self.urlhash:
+            hash_val = self.__class__.objects.generate_urlhash()
+            self.urlhash = hash_val
+            while self.__class__.objects.filter(urlhash=hash_val).exists():
+                self.urlhash = self.__class__.objects.generate_urlhash()
```
Radi85, line 51: This is always evaluated to True in case urlhash exists.
```suggestion
    def _set_unique_urlhash(self):
        if not self.urlhash:
            hash_val = self.__class__.objects.generate_urlhash()
            self.urlhash = hash_val
            while self.__class__.objects.filter(urlhash=self.hash_val).exists():
                self.urlhash = self.__class__.objects.generate_urlhash()
```

Radi85 (changes requested):

comment/tests/test_models.py
Radi85: The test helped us finding an issue with the current while statement, I already requested change with suggestion.
Here you go:

```python
    @patch('comment.managers.CommentManager.generate_urlhash')
    def test_urlhash_is_unique(self, mocked_generate_urlhash):
        mocked_generate_urlhash.side_effect = ['first_urlhash', 'first_urlhash', 'second_urlhash']
        first_comment = self.create_comment(self.content_object_1)
        self.assertEqual(first_comment.urlhash, 'first_urlhash')
        mocked_generate_urlhash.assert_called_once()
        second_comment = self.create_comment(self.content_object_1)
        self.assertEqual(second_comment.urlhash, 'second_urlhash')
```

Radi85, line 131:
```suggestion
    @patch('comment.managers.CommentManager.generate_urlhash')
    def test_urlhash_is_unique(self, mocked_generate_urlhash):
        mocked_generate_urlhash.side_effect = ['first_urlhash', 'first_urlhash', 'second_urlhash']
        first_comment = self.create_comment(self.content_object_1)
        self.assertEqual(first_comment.urlhash, 'first_urlhash')
        mocked_generate_urlhash.assert_called_once()
        second_comment = self.create_comment(self.content_object_1)
        self.assertEqual(second_comment.urlhash, 'second_urlhash')
        self.assertEqual(mocked_generate_urlhash.call_count, 3)
```

Radi85 (changes requested):

comment/models/comments.py
abhiabhi94: Both `self.urlhash` and `hash_val` have the same values. Also, if it were always true, won't the loop run indefinitenly?

Do I miss something here?

comment/tests/test_models.py
abhiabhi94: I just tested your code. It is not passing on my machine. Its giving `StopIteration` which is probably when it is not able to set the given value.

Radi85: yes bcz you didn't change while loop issue, as I suggested

abhiabhi94: Okay it passed after using your suggestion. I'm really not sure why.

comment/models/comments.py
Radi85: No thery are not, `hash_val` is const and doesn't change inside while loop, `self. urlhash` is updating inside while loop. therefore you have to test `self.urlhash` if exist. to be honest `hash_val` is redundant and has no meaning, yo could do `self.urlhash = self.__class__.objects.generate_urlhash()` directly.

abhiabhi94: Yes, I got you. A blunder on my side. Thanks for pointing this out.

comment/tests/test_models.py
Radi85: because `self.__class__.objects.filter(urlhash=hash_val).exists()` is always True, you don't update `hash_val` inside while loop.

comment/models/comments.py
Radi85: It's not a big deal, we all do this frequently

comment/tests/test_models.py
abhiabhi94: got it, this was made before you explained in the conversation thread on the suggestion on the `model.`

comment/models/comments.py
abhiabhi94: I will try to add the test for migrations now and push this in a bit.

abhiabhi94: And remember whatever happens, I love you for solving this mock thing :-p This was really annoying me for long.

Radi85: I appreciate your efforts and your contributions are really valuable, I learned a lot from you as well. Many thanks!

abhiabhi94:
I have changed the test. The running of tests for migrations is getting a bit more complex since the model class there is fake one. I tried but couldn't write a test to cover the while loop there.

There's an obvious point to import and use the same function in the migration file but that would cause issue with migration history if we change this function anytime in the future.

My logical part of the brain is too exhausted for the day to figure this test. If you happen to write please inform me as well. Please free to merge this or rebase and add code as per your wish.

Radi85 (approved):

PR #41, opened 2020-08-01
Add option to comment anonymously

### LINKED ISSUE #33: Allow anonymous posting of comments
Hey Radi.

By any chance, would you want to allow anonymous posting of comments in the application?

In case you are interested, I would suggest the possible workflow:

- Ask anonymous users to give their email alongside the comment.
- We would send an email to them with a link to verify.
- The comment would only hit the database when they verify.

Obviously, this feature would be a choice based one for a developer.

abhiabhi94:
- unauthenticated users can comment anonymously.
- email is taken alongside comment for authentication, although it is not stored.
- user can verify the comment by click on the link sent to their email.
- the email can be sent in both html and text format.
- comment only hits the database when it is authenticated.

Closes #33


abhiabhi94:
Hey @Radi85,

I had never thought that #33 would open a Pandora box. Although it's been fun, I'm currently stuck somewhere and would request you to help.

This is the current workflow.

- We add an email field to the form when the user is authenticated and the setting allows for anonymous commenting. An email is sent to the user's email with a link to verify the comment.

- The comment object is converted to dictionary(so that it can be serializable for signing) and encrypted using the `SECRET_KEY` as the key. This is then sent to the email address with a link(in the encrypted form) to verify the comment.

- When the user clicks on the email we intend to verify the comment. It can be successfully verified here and the comment can be decrypted from the link and added to the database. 

But the issue occurs when this link is clicked again or hence forth. We don't really have a method to understand this link has already been verified. Since we don't store anything in the database related to this comment, I don't really understand how we can verify the comment?

Some things that have crossed my mind are:

- Adding some sort of identifier(UUID) to the comment and adding that to the database. But this would mean creating a separate table in the database and wouldn't be a great design in my opinion.

- Another option would be to send an `id` which is obtained from the `comment` table. Suppose the id of the last comment is 56, then 57 in some form would be sent in the link. The database also needs to know that 57 can't be given and it would now give the next comment an `id` 58. This would require interacting with the database and using its incrementing function for the column `id`. I'm not really this could be a good alternative because Django > 3.0 supports non-SQL databases as well, which would make this difficult to handle for all databases.

- Another option would be to send a time stamp alongside the link, use the fact that content and time stamp for this comment is unique. This would fail if multiple people comment the same content on the same time. This could be used if we store the email, but I don't really want to do it. 

Any of your ideas would be great.

abhiabhi94:
Once you verify the current architecture and workflow, I will start adding tests and make the required changes to the javascript file.

Radi85:
Hey @abhiabhi94.

Thanks agian for your efforts!

> * We add an email field to the form when the user is authenticated and the setting allows for anonymous commenting. An email is sent to the user's email with a link to verify the comment.
> * The comment object is converted to dictionary(so that it can be serializable for signing) and encrypted using the `SECRET_KEY` as the key. This is then sent to the email address with a link(in the encrypted form) to verify the comment.
> * When the user clicks on the email we intend to verify the comment. It can be successfully verified here and the comment can be decrypted from the link and added to the database.
> 

The workflow looks reasonable and I agree with it.

> But the issue occurs when this link is clicked again or hence forth. We don't really have a method to understand this link has already been verified. Since we don't store anything in the database related to this comment, I don't really understand how we can verify the comment?

I would suggest to add this comment to the db and use its id in the verification link. However we need to add a boolean field `verified` to the comment model. This field will be always true for the comments done by authenticated users and false for anonymous till the email is verified.

abhiabhi94:
> I would suggest to add this comment to the db and use its id in the verification link. However we need to add a boolean field verified to the comment model. This field will be always true for the comments done by authenticated users and false for anonymous till the email is verified.

This would be problematic if a lot of unverified comments are kept in the database.

Radi85:
> This would be problematic if a lot of unverified comments are kept in the database.

I don't see any issue with that, very simple solution can be a script to clean the comment db on weekly basis.

abhiabhi94:
> > This would be problematic if a lot of unverified comments are kept in the database.
> 
> I don't see any issue with that, very simple solution can be a script to clean the comment db on weekly basis.

What if someone spams using a script?

Radi85:
> What if someone spams using a script?

I am not sure what that exactly means!
Spamming the script in this case, means the entire system is spammed already.

abhiabhi94:
> Spamming the script in this case, means the entire system is spammed already.

No I mean someone writes a script that actually fills the comment form and submits them. Invalid emails won't hit the database unless the comments are verified.

Radi85:
> No I mean someone writes a script that actually fills the comment form and submits them. Invalid emails won't hit the database unless the comments are verified.

I am not sure that fully uderstood your point, but I don't see any reaction between the email and the database directly? 
You send encrypted comment and id in the verification link, user hit the link, then you decrypt the comment and the id and check if they have matching in the db, That's it.

abhiabhi94:
Someone writes a script to fill the comment form that uses a bogus email address `a@a.com`. This script hits the API again and again. All such comments made using this script are now kept in the database, although they may be deleted in future using a script. This unnecessarily increases the database interaction and populates it too.

Earlier, only comments by authenticated users would hit the database.

Radi85:
> Someone writes a script to fill the comment form that uses a bogus email address `a@a.com`. This script hits the API again and again. All such comments made using this script are now kept in the database, although they may be deleted in future using a script. This unnecessarily increases the database interaction and populates it too.
> 
> Earlier, only comments by authenticated users would hit the database.

Hey @abhiabhi94,
sorry for being late, too many thing to deal with.

You are right and for this reason we should avoid any solution react with the DB, I would currently suggest to save the encrypted comment in a session that expires in 30 mins.

comment/models/comments.py
```diff
@@ -7,7 +7,7 @@
 
 
 class Comment(models.Model):
-    user = models.ForeignKey(get_user_model(), on_delete=models.CASCADE, default=None)
+    user = models.ForeignKey(get_user_model(), on_delete=models.CASCADE, blank=True, null=True)
```
Radi85, line 10: We should add email field to the comment model.

comment/forms.py
```diff
@@ -1,10 +1,36 @@
...
 class CommentForm(forms.ModelForm):
     class Meta:
         model = Comment
         fields = ('content', )
         widgets = {'content': forms.Textarea(attrs={'rows': 1})}
+
+    def __init__(self, *args, **kwargs):
+        self.request = kwargs.pop('request')
+        super().__init__(*args, **kwargs)
+        if (self.request.user.is_anonymous) and (settings.COMMENT_ALLOW_ANONYMOUS):
```
Radi85, line 17: Please remove redundant parentheses in if statements.

comment/views/comments.py
```diff
@@ -45,22 +57,41 @@ def form_valid(self, form):
         model_id = self.request.POST.get('model_id')
         model_object = get_model_obj(app_name, model_name, model_id)
         parent_id = self.request.POST.get('parent_id')
-        parent_comment = None
         if parent_id:
             parent_qs = Comment.objects.filter(id=parent_id)
             if parent_qs.exists():
                 parent_comment = parent_qs.first()
+        else:
+            parent_comment = None
```
Radi85, line 65: Please revert the code to the old state.
```suggestion
        parent_comment = None
        if parent_id:
            parent_qs = Comment.objects.filter(id=parent_id)
            if parent_qs.exists():
                parent_comment = parent_qs.first()
            
```

comment/conf/defaults.py
```diff
@@ -13,3 +14,7 @@
 COMMENT_URL_PREFIX = 'comment-'
 COMMENT_URL_SUFFIX = ''
 COMMENT_URL_ID_LENGTH = 8
+COMMENT_ALLOW_ANONYMOUS = False
+COMMENT_FROM_EMAIL = settings.EMAIL_HOST_USER   # used for sending confirmation emails
+COMMENT_CONTACT_EMAIL = COMMENT_FROM_EMAIL  # used in confirmation email
```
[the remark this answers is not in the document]
abhiabhi94: This is the default value. Suppose you want to send the email from <no-reply@a.com> but want the user to contact at <contact@a.com>. It is for this particular case.

comment/models/comments.py
abhiabhi94: Any particular reason? As a matter of personal preference, I wanted to make these comments completely anonymous and hence didn't really want to save the email.

comment/views/comments.py
abhiabhi94: In my opinion, it is better to define variables where they are used. While going through this logic, I was first confused as to why the variable `parent_comment` was given a value `None` before going I read the lines below this. The newer version seems more clear to me unless there is some other issue.

abhiabhi94:
> sorry for being late, too many thing to deal with. 

No, issues... I hope you're okay. I too, have been busy with some of my personal stuff.

> You are right and for this reason we should avoid any solution react with the DB, I would currently suggest to save the encrypted comment in a session that expires in 30 mins.

This seems legit. Let me try to implement this and see.

comment/conf/defaults.py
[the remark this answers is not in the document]
Radi85: I think I am missing something here, but what I see in the setting file is that `COMMENT_FROM_EMAIL` and `COMMENT_CONTACT_EMAIL` have the same value (email)
```
COMMENT_FROM_EMAIL = settings.EMAIL_HOST_USER  
COMMENT_CONTACT_EMAIL = COMMENT_FROM_EMAIL  # they are same
```

comment/models/comments.py
Radi85: Yes there is a good reason for that which is #43 
Email field is not only required for the anonymous comment but also for the authenticated user for the notification system later.

comment/conf/defaults.py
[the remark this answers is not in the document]
abhiabhi94: Yes you are right. I will fix this. I was testing something here and forgot to change this back.

Radi85:
@abhiabhi94 
Generally we need to look for a way to improve the UI, instead of showing the email field with the comment content field, we might request inserting the emal in a modal form when the comment is done anonymously. I believe this would be much better from user experience's view.

comment/models/comments.py
abhiabhi94: I do see the feature request as a great one. Maybe we may put something like this somewhere that anonymous comment users won't be notified?

abhiabhi94:
> Instead of showing the email field with the comment content field, we might request inserting the emal in a modal form when the comment is done anonymously. I believe this would be much better from user experience's view.

I was planning to show `Anonymous User` has the content author and content below it.

comment/models/comments.py
Radi85: why would you adopt that? when someone is commenting anonymously, still wanting to be notified for new updates in the thread. We can still implementing unsubscribe feature to remove them from emailing list.

abhiabhi94: Yes I do get your point and it is a valid one. Okay maybe when we are implementing the notification thing, we may put a check where the user can choose to be notified. If they do choose to be notified, their email can be stored then.

I don't see a point of storing the email in the current state.

comment/views/comments.py
Radi85: I believe we should always initialise the variable and then update its value. what if `parent_qs.exists()` is false? The parent_comment will be undefined.

abhiabhi94: If `parent_id` exists `parent_qs` would exist? Do i miss something here?

abhiabhi94:
> > sorry for being late, too many thing to deal with.
> 
> No, issues... I hope you're okay. I too, have been busy with some of my personal stuff.
> 
> > You are right and for this reason we should avoid any solution react with the DB, I would currently suggest to save the encrypted comment in a session that expires in 30 mins.
> 
> This seems legit. Let me try to implement this and see.

Hey, won't this restrict the user to verify the comment on the same device which they use for commenting? Is this a good option?

abhiabhi94:
> Hey, won't this restrict the user to verify the comment on the same device which they use for commenting? Is this a good option?

Forget that I wrote this. I was thinking about storing sessions on the frontend.

[Storing sessions as a file based](https://docs.djangoproject.com/en/3.1/topics/http/sessions/#using-file-based-sessions). Would this be sufficient?

comment/models/comments.py
Radi85: Saving the comment without any reference (username or email) will give people the whole chance to abuse the use of comment and the flagging system will be in this case useless. And also any one can spam the thread with dirty stuff :D Therefore the comment must be saved with username or email.

comment/views/comments.py
Radi85: This is not always true, what if the author of parent comment deleted it while another user is replying to it? The parent_id exists since is passed with comment form and parent_qs doesn't exists. To make it simple this code logic is not valid in other languages :)

Radi85:
> Forget that I wrote this. I was thinking about storing sessions on the frontend.
> 
> [Storing sessions as a file based](https://docs.djangoproject.com/en/3.1/topics/http/sessions/#using-file-based-sessions). Would this be sufficient?

I think yes. But do we really need to go with this option?

Radi85:
> I was planning to show `Anonymous User` as the content author and content below it.

Didn't get what you mean.

comment/models/comments.py
abhiabhi94: In my opinion, verifying by clicking on the link sent to their email makes spamming consistently more difficult. People can always spam a thread if they want to by creating an account.

If you see in essence, we are verifying their email account which is what mostly most websites do before account creating. Just the fact that we currently, do it before accepting every comment makes it more difficult to spam.

comment/views/comments.py
abhiabhi94: Yes, you are right.

abhiabhi94:
> > I was planning to show `Anonymous User` as the content author and content below it.

For normal comments, we show a username beside a profile picture alongside the comment. For anonymous comments, I was planning to show some avatar in place of the profile picture, anonymous user in place of the username. 


> > [Storing sessions as a file based](https://docs.djangoproject.com/en/3.1/topics/http/sessions/#using-file-based-sessions). Would this be sufficient?
> 
> I think yes. But do we really need to go with this option?

[By default, django uses the databases for storing sessions.](https://docs.djangoproject.com/en/3.1/topics/http/sessions/#enabling-sessions). The cache based one is more efficient performance wise. The file based method would be a little lesser complex to use for the end developer since they don't have to run separate migrations for this.

Radi85:
> For normal comments, we show a username beside a profile picture alongside the comment. For anonymous comments, I was planning to show some avatar in place of the profile picture, anonymous user in place of the username.

I agree. That's why I will continue gravatar implementation after this done.


> [By default, django uses the databases for storing sessions.](https://docs.djangoproject.com/en/3.1/topics/http/sessions/#enabling-sessions) The cache based one is more efficient performance wise. The file based method would be a little lesser complex to use for the end developer since they don't have to run separate migrations for this.

I would avoid this complexity, I believe many developers would avoid using the app for this reason.

comment/models/comments.py
Radi85: For this reason we should implement blocking list in the future. For the coming features, I see the email field in the comment model is a must.

abhiabhi94:
> I would avoid this complexity, I believe many developers would avoid using the app for this reason.

Okay, so how exactly would you like us to store comments before they are verified?

Radi85:
> Okay, so how exactly would you like us to store comments before they are verified?

I don't mind implementing file session as I see it the best option for now, though the extra configuration might be frustrated for some developers.

abhiabhi94:
> > Okay, so how exactly would you like us to store comments before they are verified?
> 
> I don't mind implementing file session as I see it the best option for now, though the extra configuration might be frustrated for some developers.

I agree. I was earlier didn't want to store the email. Now that we really want to keep them, I think we may probably use the combination of email and submission time to avoid the cache thing. We would just need to change the posted time's attributes and supply a default value to current time in the model.

Radi85:
> I agree. Earlier, I didn't want to store the email. Now that we really want to keep them, I think we may probably use the combination of email and submission time to avoid the cache thing. We would just need to change the posted time's attributes and supply a default value to current time in the comment model.

Okay. Again where the email will be stored?

abhiabhi94:
I won't be stored until it is verified. It will be sent in a decrypted form in the confirmation link.

Radi85:
> I won't be stored until it is verified. It will be sent in a decrypted form in the confirmation link.

That means, when the verification request is sent, the email will bealready  in the encrypted data and the comment will be saved with this email, is that right?

abhiabhi94:
yes

comment/managers/comments.py
```diff
@@ -53,3 +55,25 @@ def generate_urlhash(self):
...
+    def get_dict(self, instance):
+        opts = instance._meta
+        data = {}
+        for f in chain(opts.concrete_fields, opts.private_fields):
+            if getattr(f, 'ct_field', None):
+                continue
+            data[f.name] = f.value_from_object(instance)
+        for f in opts.many_to_many:
+            data[f.name] = [i.id for i in f.value_from_object(instance)]
+        return data
```
abhiabhi94, line 76: I'll remove this method, this isn't used now.

`content_type` and `object_id` can't be injected directly to the `Comment` model. We also need `app_name`, `model_name`, `model_id` before obtaining the `content_object`, I thought it would be better to personally form the dictionary in the view method.

This will be removed in the final commit

test/example/post/models.py
```diff
@@ -34,7 +34,7 @@ def save(self, *args, **kwargs):
                 _title = _title[:45]
             unique_slug = self.slug = slugify(_title)
             count = 1
-            while Post.objects.filter(slug=unique_slug).exists():
```
abhiabhi94, line 37: This was wrong. Post is obviously not defined inside itself. Maybe because it never got executed, this was missed.

comment/mixins.py
```diff
@@ -0,0 +1,17 @@
...
+from django.shortcuts import get_object_or_404
+
+from comment.models import Comment
+
+
+class AnonymousPermissionMixin:
+    def has_permission(self, request, view):
+        self.comment = get_object_or_404(Comment, pk=self.kwargs.get('pk'))
+        if not self.comment.user:
+            raise PermissionDenied
```
Radi85, line 12: I believe this is doing same what `LoginRequiredMixin` is doing.

comment/views/comments.py
```diff
@@ -45,22 +58,48 @@ def form_valid(self, form):
...
+                'msg': _("""We've have sent a verification link to your email. The comment will be displayed\
+                    after your verification.""")
+            })
+        comment.save()
+        self.comment = comment
         return self.render_to_response(self.get_context_data())
 
 
-class UpdateComment(BaseCommentView):
+class UpdateComment(BaseCommentMixin, AnonymousPermissionMixin, BaseCommentView):
```
Radi85, line 102: `AnonymousPermissionMixin` is useless here. dispatch method is already performing the check against comment owner and anonymous comment will never have owner.

comment/views/comments.py
```diff
@@ -71,20 +110,20 @@ def dispatch(self, request, *args, **kwargs):
...
+        form = CommentForm(request.POST, instance=self.comment, request=self.request)
         context = self.get_context_data()
         if form.is_valid():
             form.save()
             context['comment'] = self.comment
             return render(request, 'comment/comments/comment_content.html', context)
 
 
-class DeleteComment(BaseCommentView):
+class DeleteComment(BaseCommentMixin, AnonymousPermissionMixin, BaseCommentView):
```
Radi85, line 126: Same case for update class. I will move these permission checks from dispatch method to the separated permission class after this PR is merged

comment/views/comments.py
```diff
@@ -105,3 +144,29 @@ def post(self, request, *args, **kwargs):
...
+            app_name = tmp_comment_dict.pop('app_name')
+            tmp_comment_dict.update(
+                {
+                    'content_object': get_model_obj(app_name, model_name, model_id)
+                }
+            )
+            tmp_comment = Comment(**tmp_comment_dict)
+        except (ValueError, KeyError, signing.BadSignature):
+            raise Http404
+        if Comment.objects.comment_exists(tmp_comment):
```
Radi85, line 165: is there way to make the confirmation link valid for one time only?

test/example/post/models.py
Radi85: I don't think this is wrong and Post is actually defined inside itself :). if your statement is true, I would not be able to create a post.

Radi85:
> anonymous-comment

The user shall be kept in the same view after posting the comment, redirecting to the home page is not user friendly 😄 
we also should look at improving the UI for anonymous case

comment/apps.py
```diff
@@ -8,19 +9,19 @@ def create_permission_groups(sender, **kwargs):
     from comment.models import Comment
 
     comment_ct = ContentType.objects.get_for_model(Comment)
-    delete_comment_perm, _ = Permission.objects.get_or_create(
+    delete_comment_perm, __ = Permission.objects.get_or_create(
```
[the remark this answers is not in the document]
abhiabhi94: `_` is used as an alias for `gettext_lazy` for the verbose name attribute. This was used as the same variable name elsewhere in the project, so keep things consistent I did this.
I keep adding the translation capabilities wherever I see something while debugging or going through the code. It just would make the process easier in the future.

comment/mixins.py
abhiabhi94: `LoginRequiredMixin` checks whether the user is logged in or not. This checks whether the comment has an author or not.

comment/views/comments.py
abhiabhi94: Yes the line `if Comment.objects.comment_exists(tmp_comment):` does the same thing as you ask.

comment/conf/defaults.py
```diff
@@ -13,3 +14,8 @@
 COMMENT_URL_PREFIX = 'comment-'
 COMMENT_URL_SUFFIX = ''
 COMMENT_URL_ID_LENGTH = 8
+COMMENT_ALLOW_ANONYMOUS = False
+if getattr(settings, 'COMMENT_ALLOW_ANONYMOUS', COMMENT_ALLOW_ANONYMOUS):
+    COMMENT_FROM_EMAIL = settings.EMAIL_HOST_USER   # used for sending confirmation emails
+    COMMENT_CONTACT_EMAIL = COMMENT_FROM_EMAIL  # used for contact address in confirmation emails
```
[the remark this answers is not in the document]
abhiabhi94: > > anonymous-comment
> 
> The user shall be kept in the same view after posting the comment, redirecting to the home page is not user friendly smile

They aren't redirected. You replace the `comment` html container with the data received from the backend. Since I return a `JsonResponse` currently and I have not made any changes to `javascript`, it might seem like a redirect. Since the response is not handled anyway right now at the frontend, the container appears to be empty and hence a redirect.


> we also should look at improving the UI for anonymous case

Yes, any suggestions would be great.

abhiabhi94:
> > anonymous-comment
> 
> The user shall be kept in the same view after posting the comment, redirecting to the home page is not user friendly smile

They aren't redirected. You replace the `comment` html container with the data received from the backend. Since I return a `JsonResponse` currently and I have not made any changes to `javascript`, it might seem like a redirect. Since the response is not handled anyway right now at the frontend, the container appears to be empty and hence a redirect.


> we also should look at improving the UI for anonymous case

Yes, any suggestions would be great.

comment/views/comments.py
abhiabhi94: Yes you are right, I didn't see the dispatch method here. I will remove the mixin.

comment/templatetags/comment_tags.py
```diff
@@ -24,9 +24,8 @@ def get_app_name(obj):
 def get_profile_url(obj):
     """ returns profile url of user """
     content_type = get_profile_content_type()
-    if not content_type:
+    if (not content_type) or (not obj.user):
```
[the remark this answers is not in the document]
abhiabhi94:
```python

not a or not b
```
This will open the room for operator precedence. Using brackets makes the operation more predictable.

comment/conf/defaults.py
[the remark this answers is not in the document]
abhiabhi94: Okay, I'm a bit confused here. Do we define the default values inside this file or the actual values of these variables here.

As far as I see, we are defining the default values here. If that is the case, the default values of the variables `COMMENT_FROM_EMAIL` and `COMMENT_CONTACT_EMAIL` are same unless specified otherwise by the user in their settings.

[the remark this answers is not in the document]
Radi85: so in this case it should be like this:
```python
COMMENT_CONTACT_EMAIL = getattr(settings, 'COMMENT_CONTACT_EMAIL', COMMENT_FROM_EMAIL)
```

comment/templatetags/comment_tags.py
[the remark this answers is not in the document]
Radi85:
```python
if not content_type or not obj.user:  # what's wrong with this?
```

comment/mixins.py
Radi85: Yes. what I meant here that the chack you mentioned is already implemented in dispatch method and `AnonymousPermissionMixin` no longer needed

comment/conf/defaults.py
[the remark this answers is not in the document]
abhiabhi94: Okay, this is what I thought.

Shouldn't the other variables defined inside this file like
```python

COMMENT_FLAG_REASONS = [
    (1, _('Spam | Exists only to promote a service')),
    (2, _('Abusive | Intended at promoting hatred')),
]
```
be defined the same way as:

```python

COMMENT_FLAG_REASONS = getattr(settings, 'COMMENT_FLAG_REASONS', [
    (1, _('Spam | Exists only to promote a service')),
    (2, _('Abusive | Intended at promoting hatred')),
])

```

comment/views/comments.py
Radi85: I know, I asked bcz I wanted to get rid of this line and make sure that the link is one time valid. However, I can live with this check.

comment/mixins.py
abhiabhi94: Yes, I agree.

comment/views/comments.py
abhiabhi94: Any other solution would mean we would have to use the database or caches etc which would make things more complex. So, till we find a more efficient way, we will live have to live with this :-p

comment/templatetags/comment_tags.py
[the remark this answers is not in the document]
abhiabhi94: nothing just that in this case operator precedence rule take over. `not` is evaluated first before the `or` evaluation and bla bla bla.

I just don't want to remember this information and have predictability in my operations. If you really don't like the cute parenthesis :-p, I'll remove them

comment/views/comments.py
```diff
@@ -45,22 +58,48 @@ def form_valid(self, form):
...
+            site = get_current_site(self.request)
+            key = signing.dumps(dict_comment, compress=True)
+            send_email_confirmation_request(comment, email, key, site)
+            return JsonResponse({
+                'msg': _("""We've have sent a verification link to your email. The comment will be displayed\
+                    after your verification.""")
+            })
+        comment.save()
+        self.comment = comment
         return self.render_to_response(self.get_context_data())
```
Radi85, line 99:
```suggestion
        context = self.get_context_data()
        if settings.COMMENT_ALLOW_ANONYMOUS and self.request.user.is_anonymous:
            # send response, please verify your email to post this comment.
            site = get_current_site(self.request)
            key = signing.dumps(dict_comment, compress=True)
            send_email_confirmation_request(comment, email, key, site)
               context['msg']: _("""We've have sent a verification link to your email. The comment will be displayed\
                    after your verification.""")
        else:
            comment.save()
            self.comment = comment
        return self.render_to_response(context)
```

Radi85:
> They aren't redirected. You replace the `comment` html container with the data received from the backend. Since I return a `JsonResponse` currently and I have not made any changes to `javascript`, it might seem like a redirect. The response is not handled anyway right now at the frontend, hence the container appears to be empty which might have seemed as a redirect in th e picture.

The response might be updated as show here -> https://github.com/Radi85/Comment/pull/41#discussion_r469894333

> Yes, any suggestions would be great.

I was thinking of using modal form for the email in case anonymous user is posting.

comment/views/comments.py
abhiabhi94: Yes, I'' fix this part, it's just I first wanted to get this working. Also `self.comment` won't be defined in the true part of this branch which would create an issue inside this function:

```python
def get_template_names(self):
        if self.comment.is_parent:
            return ['comment/comments/base.html']
        else:
            return ['comment/comments/child_comment.html']
```
No issues, I have looked more closely into this right now. I'll try to figure this out.

comment/conf/defaults.py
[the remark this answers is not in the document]
Radi85: Hmmm, I got what you mean. The developer settings will override the value of any existing attribute in the default settings.

so I would prefer going this way:

```python
COMMENT_FLAG_REASONS = [
    (1, _('Spam | Exists only to promote a service')),
    (2, _('Abusive | Intended at promoting hatred')),
]
```

abhiabhi94:
> I was thinking of using modal form for the email in case anonymous user is posting.

Okay, the placeholder text could be saved into the model attribute as a help text and title will be added dynamically here. Title is important for screen reader to have access to the container, which is important for a lot of users.

Since editing would be required anywhich way here, I wrote this code. Also, I don't understand how will the User Interface(UI) be better in this case?

comment/templatetags/comment_tags.py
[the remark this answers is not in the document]
Radi85: 😄 I hate parenthesis in if statement. I also don't like to see python code written in JS style 🤮 

```python
for i in range(len(itms)):
    give_me_an_item = items[i]  # 🤮 
```

comment/conf/defaults.py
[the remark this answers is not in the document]
abhiabhi94: Yes. If you see the GIF attached closely the contact email is `contact@domain`(picked from developer settings) which has overriden the default values
That means I don't have to make any changes as suggested earlier in this discussion?

comment/views/comments.py
Radi85: I think just update if statement to return base.html if user anonymous also 
`if not self.request.user.is_authenticated or self.comment.is_parent`the first part will be evaluated first before `self.comment.is_parent` is called

abhiabhi94: Okay, seems legit.

comment/conf/defaults.py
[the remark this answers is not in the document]
Radi85: You are right. It's confusing. 😄

comment/views/comments.py
abhiabhi94: Also, after this is fixed, I will look to add an option where `ctrl + enter` submits the comment. It will be more time saving thing instead of going to the touchpad/mouse and clicking on the button.

Radi85:
> Okay, the placeholder text could be saved into the model attribute as a help text and title will be added automatically here. Title is important for screen reader to have access to the container, which is important for a lot of users.
> 
> Since editing would be required any which way here, I wrote this code. Also, I don't understand how will the User Interface(UI) be better in this case?

It would be much better that showing 2 fields in the parent section and in the replies. I would keep the comment content field as it's shown for the authenticated user and when anonymous try to post a comment, they should be prompted with the modal to insert their email for proceeding. You may leave this part and I will implement it later.

abhiabhi94:
> > Okay, the placeholder text could be saved into the model attribute as a help text and title will be added automatically here. Title is important for screen reader to have access to the container, which is important for a lot of users.
> > Since editing would be required any which way here, I wrote this code. Also, I don't understand how will the User Interface(UI) be better in this case?
> 
> It would be much better than showing 2 fields in the parent section and in the replies. I would keep the comment content field as it's shown for the authenticated user and when anonymous try to post a comment, they should be prompted with the modal to insert their email for proceeding. You may leave this part and I will implement it later.

Okay I mixed modal and model :-p Yes this sounds better.

comment/api/serializers.py
```diff
@@ -98,28 +99,52 @@ def get_reactions(obj):
...
+                raise serializers.ValidationError(
+                    _('Email is required for posting anonymous comments'),
+                    code='invalid'
+                    )
+
+        return value.strip().lower()
+
     def create(self, validated_data):
+        # check if anonymous posting is allowed and user is unauthenticated
+        # send email by creating a key from the comment dict.
```
abhiabhi94, line 142: Can we handle this the same way (I mean temporart `object` and `dict` creation) so that while confirming the comment, the key can be handled by the function `get_comment_from_dict` in `utils.py`?

I do understand that keys are different. I just wanted your take to know if something better is possible here.

comment/utils.py
```diff
@@ -88,6 +101,11 @@ def get_comment_context_data(request, model_object=None, comments_per_page=0):
 
     allowed_flags = getattr(settings, 'COMMENT_FLAGS_ALLOWED', 0)
     oauth = request.POST.get('oauth') or request.GET.get('oauth')
+    if oauth == 'true':
+        oauth = True
+    else:
+        oauth = False
+    is_anonymous_allowed = getattr(settings, 'COMMENT_ALLOW_ANONYMOUS', False)
```
abhiabhi94, line 108: Since `oauth` was coming from an HTTP request, it was a string. It would always be true in the template.
I think this was unintended.

comment/models/comments.py
```diff
@@ -49,10 +51,21 @@ def _set_unique_urlhash(self):
...
     def save(self, *args, **kwargs):
         self._set_unique_urlhash()
+        self._set_email()
         super(Comment, self).save(*args, **kwargs)
 
+    def get_absolute_url(self):
+        page_url = self.content_object.get_absolute_url()
+        if not page_url.endswith('/'):
+            page_url += '/'
+        return page_url + '#' + self.urlhash
```
abhiabhi94, line 67: This is also not correct. There needs to be some consideration for pagination but since the comments per page is retrieved from the template, I currently, don't understand how this can be fixed here.

comment/api/serializers.py
Radi85: Yes. It's the best using this function `get_comment_from_key`.

comment/forms.py
```diff
@@ -1,10 +1,40 @@
...
 
 from comment.models import Comment
+from comment.conf import settings
 
 
 class CommentForm(forms.ModelForm):
     class Meta:
         model = Comment
-        fields = ('content', )
+        fields = ('content', 'email')
```
Radi85, line 11: The old approach seems to be better. I don't like unnecessary hidden field. Construct the field on `if self.request.user.is_anonymous and settings.COMMENT_ALLOW_ANONYMOUS` is true.

comment/models/comments.py
Radi85: Won't `self.urlhash` take you to the comment regarless the pagination? If the answer NO, then a dirty solution is that we may be able to know in which page the comment is based on posted time.

comment/utils.py
```diff
@@ -88,6 +101,11 @@ def get_comment_context_data(request, model_object=None, comments_per_page=0):
 
     allowed_flags = getattr(settings, 'COMMENT_FLAGS_ALLOWED', 0)
     oauth = request.POST.get('oauth') or request.GET.get('oauth')
+    if oauth == 'true':
+        oauth = True
+    else:
+        oauth = False
```
Radi85, line 107: You are right. I would suggest the following:
```suggestion
    if oauth and oauth.upper() == 'TRUE':
        oauth = True
    else:
        oauth = False
```

comment/forms.py
abhiabhi94: Yes, seems legit.

comment/models/comments.py
abhiabhi94: > Won't self.urlhash take you to the comment regarless the pagination

No, the comment won't be present on the page due to pagination.

> a dirty solution is that we may be able to know in which page the comment is based on posted time

As far as I remember, comments per page is obtained from the template. Without it, how can we know the page number?

Radi85: Moving pagination vars to settings?

comment/api/serializers.py
abhiabhi94: The keys for the dictionary required to form the link(key) is different here.

comment/models/comments.py
abhiabhi94: yes, this can be one of the solutions.

Radi85: You are right. But this won't significantly affect the app.

comment/api/serializers.py
Radi85: I am not sure what you mean.

abhiabhi94: In APIs, we have `model_id` and `model_name` from the HTTP requests while in the other one we use `model_name`, `model_app` and `app_name`. So, the method used for object(comment) creation is also different.

Radi85: I got confused, there is no model_app variable I think you meant model_id.
Cannot we get the app_name from model_type (this is the name of model_name in the api)?

abhiabhi94: > Cannot we get the app_name from model_type (this is the name of model_name in the api)

i am not sure i completely understood what you meant.

Radi85: What I meant is that we can get the app name from the model name. However, I see that I am using 2 different ways to create a comment so it would be cleaner if we can use only one method to create a comment.

abhiabhi94: can the same model name not be used in two different applications?

Radi85: Exactly. This will cause an issue. Therefore the app name must be provided in the api as well.

Radi85: It can be optional and we can perform the check if it's needed, then the delveloper will be asked to provide it

Radi85: I already opened this #46 to fix it.

abhiabhi94: A one time check won't be sufficient. the check will have to be performed at every request. wouldn't it be better to have this information in the API request?

abhiabhi94: okay, i will add this field in the API.

Radi85: > A one time check won't be sufficient. the check will have to be performed at every request. wouldn't it be better to have this information in the API request?

Yes. If the checks fails, we ask the user to provide the app name in url query string. However, you might make it required now and later we look for the optional choice.

comment/api/serializers.py
```diff
@@ -99,31 +101,69 @@ def get_reactions(obj):
...
+            'email': email,
+            'posted': str(time_posted),
+            'app_name': self.app_name,
+            'model_name': self.model_name,
+            'model_id': self.model_id,
+            'parent': parent_id
+        }
+        if settings.COMMENT_ALLOW_ANONYMOUS and (not user):
+            response_msg = process_anonymous_commenting(request, comment, dict_comment, api=True)
+            self.context['msg'] = response_msg
```
abhiabhi94, line 163: what should I return here. I can obviously override the `post` method in the view of `generics.CreateAPIVIew` but I thought there may be an easier way to do this. Any suggestion from you would be great.

abhiabhi94:
i have moved pagination to settings. only the response for anonymous commenting is left, otherwise most of the work is almost done.
if you approve everything else, i will start adding tests and fix the old ones.

comment/models/comments.py
```diff
@@ -49,10 +53,29 @@ def _set_unique_urlhash(self):
...
+            page_url += '/'
+        comments_per_page = settings.COMMENT_PER_PAGE
+        if comments_per_page:
+            qs_all_parents = self.__class__.objects.filter_parents_by_object(
+                self.content_object, include_flagged=is_comment_moderator(request.user)
+                )
+            position = qs_all_parents.filter(posted__gte=self.posted).count() + 1
+            if position > comments_per_page:
+                page_url += '?page=' + str((position // comments_per_page) + 1)
+        return page_url + '#' + self.urlhash
```
abhiabhi94, line 77: although this works well, it is currently not used if the comment is hidden as a reply. the url is correct but since the element is hidden, it doesn't show.

Radi85: Nice work! This is not a big deal. Since this url leads to the parten comment, the user will be aware that their comment in the replies section.

abhiabhi94: no it doesn't lead anywhere because the id is currently not present(hidden).

comment/api/serializers.py
Radi85: What about None?

Radi85: What about None?

comment/models/comments.py
abhiabhi94: this also assumes the comments are sorted date wise although we may or the user may choose to sort them based on other parameter for e.g. replies or likes etc

comment/api/permissions.py
```diff
@@ -11,8 +11,11 @@
 class IsOwnerOrReadOnly(permissions.BasePermission):
     """
     Custom permission to only allow owner of an object to edit it.
+    Anonymous objects can't be edited.
     """
     def has_object_permission(self, request, view, obj):
+        if not obj.user:    # anonymous comment
```
Radi85, line 17: This statement is needless here because `obj.user == request.user` is always false for anonymous comment.

comment/api/serializers.py
abhiabhi94: None is obviously being returned here but it creates an issue with the `generics.CreateAPIView` inherited in the `view` function. It expects a model instance to be returned from the `create` function otherwise it shows an error.

comment/models/comments.py
Radi85: Then we should make the url always leads to the parent comment. This will make it easier.

comment/api/permissions.py
abhiabhi94: You are right. I had just added this so that this gets executed before any check and the operation happens a bit faster maybe.

Radi85: Since we have no issue with the performance, it would be much better to keep the code cleaner.

comment/models/comments.py
abhiabhi94: this can be a solution for the time being but all of this conjecture is based when comments are sorted in a definite order. I see currently we may add this option a template tag to include order if the developer want.

comment/api/serializers.py
Radi85: So I think we should use a different view for creating anonymous comment. I belive this would be cleaner since we will separate the logic.

comment/api/permissions.py
abhiabhi94: okay, seems legit. will remove it.

comment/api/serializers.py
Radi85: it just came to my mind, what about returning the comment instance, without saving it to db?

abhiabhi94: > it just came to my mind, what about returning the comment instance, without saving it to db?

we may do this alongside the message that an email has been sent bla bla...

comment/models/comments.py
Radi85: Yeah this can be implemented in a separate ticket. I would avoid passing variables to template tags onwards since the settings is available to us without hitting the root settings on every use.

abhiabhi94: in an ideal world i would like to highlight the container just as github does while moving to a discuss node when an appropriate url is entered.

abhiabhi94: okay, so i have basically copied their CSS for highlighting the target element. since, i am almost unskilled to judge the correct colour you may pick and choose one if you wish to change. i will commit the changes when they are ready

abhiabhi94:
I'm refactoring some tests. 

Just for general trivia: do you follow Test Driven Development? If yes, how do you start? If no, then why?

Just going through the test, it seemed you didn't enjoy writing tests :-p, although i may be wrong. There were a couple of tests where I needed to see the flow of the whole test to fix something or change something if things had changed.

I'm sorry this is taking a bit longer than it should have. I will commit the changes when they are ready.

Radi85:
> @Radi85 Currently, I'm refactoring some(read a lot) tests.
> 
> Just for general trivia: do you follow Test Driven Development? If yes, how do you start? If no, then why?

No I don't bcz I never tried it but I would like to. Do you? if yes, why? :D 

> Just going through the test, it seemed you didn't enjoy writing tests :-p, although i may be wrong. There were a couple of tests where I needed to see the flow of the whole test to fix something or change something if things had changed.

I do enjoy writing tests. Can you give an example?

> I'm sorry this is taking a bit longer than probably it should have. I will commit the changes when they are ready.

No worries, I am not in hurry.

abhiabhi94:
> No I don't bcz I never tried it but I would like to. Do you? if yes, why? :D

I have tried to do in a couple of my personal project. I don't have a lot of "professional" experience, but I am seeing this as a prerequisite in some job descriptions during my job hunt these days.

Personally, i liked the idea. As an example if you think of writing a function for leap year, if you write the test cases for the following case:

- when the input is not an integer.

- when it is not a multiple of 4.
- when it is a multiple of 100 but not 400.
- success case.

In my opinion, this helps in helping figure out what exact code to write and all code written is tested. But this can be done only when the processes are pretty much discussed in detailed down to micro level.

<br>

> I do enjoy writing tests. Can you give an example of the issue yo have mentioned?

If you see the [diff](https://github.com/Radi85/Comment/pull/47/files#diff-d8b1128b416497d5917ffeecde91187fL177) here, the url parameters were basically passed as raw strings everywhere and I had to change a lot of places after the change in parameter name. Hence, I wrote the function `get_url` there. Also, I will point out a couple of place where I have refactored when I push in the current changes.

no issues, just that fixing the other tests has been a kind of cumbersome or frustrating if that is the word i'm looking for. I have refactored most of them so that it is easier in the future.

Radi85:
> Also, from what I have seen this application should be more popular than it currently it based on the features that it has as opposed to its contemporaries. Maybe we deploying it somewhere and giving the user's option to test the app might help. I have seen a couple of open-source applications which use some sort of deployment on some websites that allow to host such applications for free. although, I can't remember anyone on right now.

This is a good idea, I search about this once I get a time again. I am done for today 😫

PR #47, opened 2020-08-16
Add app name parameter to REST API

### LINKED ISSUE #46: Fix multiple model name issue by providing app name in the API calls.
2 different apps might share the same model name and this will casue an issue creating a cooment by the API call.

Solution:
Provide the app name in the API call.

Blocking #41

abhiabhi94:
- modify id to model_id -> maintains consistency throughout the project.
- modify type to model_name -> maintaining consistency.

Refactor some unittests and add a few for the additional code.

Fixes #46


Radi85:
Thanks a lot @abhiabhi94

abhiabhi94:
[ContentTypePermission](https://github.com/Radi85/Comment/pull/47/files#diff-0540a1e8ec78f760a5bc8188a3431859L24) seems more like a mixin than a permission. This could also be used in the django views for form validation.

Also, I forgot to remove the `create_by_model_type` method in `CommentManager`, it is no longer required I think.

I will probably do this both in #41 or do you want me to do them one separately?

Radi85:
> I will probably do this both in #41 or do you want me to do them separately?

I will do this in a separate refactor commit.

abhiabhi94:
> > I will probably do this both in #41 or do you want me to do them separately?
> 
> I will do this in a separate refactor commit.

I have made the changes for the managers thing but mixins will be a bit complex so it may be handled after some time.

PR #49, opened 2020-08-20
refactor: use vanilla JS instead of JQuery

Radi85:


comment/static/js/comment.js
```diff
@@ -1,320 +1,386 @@
...
+            });
+            // remove pagination query when posting a new comment
+            let uri = window.location.toString();
+            if (uri.indexOf("?") > 0) {
+                let clean_uri = uri.substring(0, uri.indexOf("?"));
+                window.history.replaceState({}, document.title, clean_uri);
+            }
+        }).catch(error => {
+            console.log(error);
+            alert("Unable to post your comment!, please try again");
```
abhiabhi94, line 145: rather than using this, we may use the info function defined [here](https://github.com/Radi85/Comment/pull/49/files#diff-a13a7f5b55d01dc86d2535026a68b770R383)

abhiabhi94: also, maybe somehow we can display the error phase from the backend here so that translated strings are shown here.

abhiabhi94:
get work radico, the css part could have been tricky unless you love playing with it :-p

i believe there's a need to update the documentation as well alongside this change.

PR #52, opened 2020-08-23
Remove comment per page parameter from the frontend.

abhiabhi94:
- since the parameter is directly taken from settings, it is no longer required here.
- update documentation accordingly.


abhiabhi94:
With some many parameters in settings, I think we would have a separate section in the documentation regarding them.

PR #56, opened 2020-08-30
Fix AJAX Mixing as per django-3.1 release

abhiabhi94:
the function `request.is_ajax` is removed in `django` 3.1 according to the release text https://docs.djangoproject.com/en/3.1/releases/3.1/#id2


abhiabhi94:
this was opened due to some github bug maybe, already covered in #55

PR #58, opened 2020-09-03
Add rest framework to installed apps

abhiabhi94:
I believe this was mistakenly removed in an earlier commit.


abhiabhi94:
Also, while writing a `mixin` that would be useful for both django and rest framework, I see we are creating a comment in django when parent id passed with a wrong id of parent comment(suppose 100) and return a bad request in rest framework API.

I think this was maybe unintended. To maintain consistency, I'm returning a bad request for both the APIs from the `mixin` now. Is it okay?

Radi85:
This covered in #67

PR #59, opened 2020-09-03
Add rest framework to installed apps

abhiabhi94:
I believe this was mistakenly removed in an earlier commit.


abhiabhi94:
already covered under #58

Issue #43, opened 2020-08-08
Notification message via E-Mail


frruit:
**Notification via E-Mail**
This is currently requested by my customer and i was wondering if this feature is maybe already in planning within the library itself. If not then I can maybe support. It would be great in that case to get a proposal from you how this should be done correctly

**Describe the solution you'd like**
* First version shall have a settings parameter like COMMENT_NOTIFICATION = True which is by default set to False. It should use the django send_mail functionalities if they are configured and in use (EMAIL_HOST, ...)
* By default every new comment entry should send out a mail to all people that mark the thread as a follower. Everybody who creates a new entry, a reply or a like/unlike action shall be set automatically as a follower of the thread
* a further boolean field should be available to follow or unfollow the thread within the UI
* When a new action occures within the thread then a pretty information mail shall be send out to the thread followers

What do you about this feature request?

Hope to hear soon from you. CYA

Radi85:
This feature is actually on my list after finishing #33 but with lesser details than you have mentioned, i.e. I didn't think of implementing thread's follower but it seems to be a nice feature alongside with email notifcation.

I would like to have this feature implemented including all the points you have mentioned and since you are aware of the whole workflow here, I will be looking forward to your PR :)

Radi85:
@abhiabhi94 
I am working on this feature and I hope I will get enough time to finish it 😄 
I have so far added a new model, I think it's called `Follower` to keep tracking of the subscribers of a comment. Just wanted to share the thoughts, any idea will be welcome!

abhiabhi94:
What all features are you considering in the first version of this?

Radi85:
- Following a parent comments and the main thread.
- Email notifications.

abhiabhi94:
As this application becomes more and more complex on the front-end part, I see a reason why `django-comments-xtd` is using `react` :laughing: 

That apart, why are you using another model for subscribers? Can't it be a `many to many` field with `User` model.
`

Radi85:
> As this application becomes more and more complex on the front-end part, I see a reason why `django-comments-xtd` is using `react` 😆

That is a nice solution. In fact we work on 2 different views for the same functionality (Django typical view and API view) I was thinking of using the API for our frontend as well and get rid of django typical views. However, this can be for later if we see this is really needed.
> 
> That apart, why are you using another model for subscribers? Can't it be a `many to many` field with `User` model.
> `

[...]

PR #60, opened 2020-09-05
Add mixin to test content type and parent_id parameters from the URL

abhiabhi94:
- this works for both django and DRF.
- remove permissions that worked as mixin earlier for these parameters in drf, also corrects status code for wrong request to 400 from 403 in drf.
- add a module for testing mixin.
- add some other utility functions.

### DIFF
--- a/comment/api/permissions.py
+++ b/comment/api/permissions.py
@@ -1,10 +1,6 @@
-from django.contrib.contenttypes.models import ContentType
-from django.utils.translation import gettext_lazy as _
-
 from rest_framework import permissions
 
 from comment.conf import settings
-from comment.models import Comment
 from comment.utils import is_comment_admin, is_comment_moderator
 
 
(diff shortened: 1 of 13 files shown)


Radi85:
Hi @abhiabhi94,

I am actually working on the same issue :( I didn't know that you are doing the same. I didn't finish yet but I think I will merge this and rebase my work on it and refactor both codes accordingly. Please consider opnening a new issue before start wroking on it :)

abhiabhi94:
> I am actually working on the same issue :( I didn't know that you are doing the same. I didn't finish yet but I think I will merge this and rebase my work on it and refactor both codes accordingly. Please consider opnening a new issue before start wroking on it :)

Yes surely from now on will keep this in mind. If I remember I had told I will factor this after the closing of anonymous comments but you're right making an issue would be more apt.

abhiabhi94:
Also, is there something that you would want me to work upon currently, apart from adding documentation for testing sending of email and other changes to documentation.

Also, I'll try to refactor and move all functions receiving signals to a separate module `listeners.py` if i get the time today.

Radi85:
> Also, is there something that you would want me to work upon currently?

What about #43?

> Also, I'll try to refactor and move all functions receiving signals to a separate module `listeners.py` if i get the time today.

I was thinking of this but didn't have time. Please name the module `signals.py` instead.

> I have adding documentation for testing sending of email and other changes to documentation, in the pipeline after that.

Great!

I was thinking of releasing a new version by tomorrow, I hope I will have enough time for that, otherwise would be next week.
PS: #43 should be in the next release not this one.

PR #67, opened 2020-09-08
refactor: change mixins to validators

Radi85:
This will change content type and parent id mixins to validator classes.
Fix bug in the method of changing flag state.

### DIFF
--- a/comment/exceptions.py
+++ b/comment/exceptions.py
@@ -1,0 +1,14 @@
+from rest_framework.exceptions import APIException
+from rest_framework.status import HTTP_400_BAD_REQUEST
+
+
+class CommentBadRequest(APIException):
+    status_code = HTTP_400_BAD_REQUEST
+    default_detail = 'Bad Request'
+
+    def __init__(self, detail=None, status_code=None):
+        if status_code:
+            self.status_code = status_code
+        if not detail:
+            detail = self.default_detail
+        self.detail = detail
--- a/comment/mixins.py
+++ b/comment/mixins.py
@@ -1,113 +1,103 @@
-from django.contrib.auth.mixins import LoginRequiredMixin
-from django.contrib.contenttypes.models import ContentType
+import abc
+
+from django.contrib.auth.mixins import LoginRequiredMixin, AccessMixin
+from django.core.exceptions import ImproperlyConfigured
+from django.http import HttpResponseBadRequest, HttpResponseForbidden
 from django.utils.translation import gettext_lazy as _
 
-from comment.models import Comment
-from comment.utils import get_data_for_request, get_response_for_bad_request
-
-
-class CommentUpdateMixin(LoginRequiredMixin):
-    pass
+from comment.conf import settings
+from comment.utils import is_comment_admin, is_comment_moderator
+from comment.validators import ValidatorMixin
 
 
 class AJAXRequiredMixin:
     def dispatch(self, request, *args, **kwargs):
         if not request.META.get('HTTP_X_REQUESTED_WITH', None) == 'XMLHttpRequest':
-            return get_response_for_bad_request(why='Only AJAX request are allowed')
+            return HttpResponseBadRequest(_('Only AJAX request are allowed'))
         return super().dispatch(request, *args, **kwargs)
 
 
-class ContentTypeMixin:
-    """
-    When using this with DRF intialize `api=True` in the class.
-    Used for provinding validation of query parameters in request to match
-    a valid ContentType. Upon successful validation, `model_name`, `model_id`, `app_name`
-    can be accessed directly as class attributes.
-    """
-    api = False
-    error = ''
+class BasePermission(AJAXRequiredMixin):
+    def has_permission(self, request):
+        return True
 
-    def _set_model_name_or_error(self, val):
-        if not val:
-            self.error = "model name must be provided"
-            return
-        self.model_name = val
+    def has_object_permission(self, request, obj):
+        return True
 
-    def _set_model_id_or_error(self, val):
-        if not val:
-            self.error = "model id must be provided"
-            return
-        self.model_id = val
 
-    def _set_app_name_or_error(self, val):
-        if not val:
-            self.error = 'app name must be provided'
-            return
+class BaseCommentMixin(LoginRequiredMixin, BasePermission):
+    pass
 
-        if not ContentType.objects.filter(app_label=val).exists():
-            self.error = f'{val} is NOT a valid app name'
-            return
-        self.app_name = val
 
-    def _set_error(self):
-        try:
-            model_name = self.model_name.lower()
-            ct = ContentType.objects.get(model=model_name).model_class()
-            model_class = ct.objects.filter(id=self.model_id)
-            if not model_class.exists() and model_class.count() != 1:
-                self.error = f"{self.model_id} is NOT a valid model id for the model {self.model_name}"
-        except ContentType.DoesNotExist:
-            self.error = f"{self.model_name} is NOT a valid model name"
-        except ValueError:
-            self.error = f"model id must be an integer, {self.model_id} is NOT"
+class CanCreateMixin(BasePermission, AccessMixin, ValidatorMixin):
+    def has_permission(self, request):
+        return request.user.is_authenticated or settings.COMMENT_ALLOW_ANONYMOUS
 
     def dispatch(self, request, *args, **kwargs):
-        data = get_data_for_request(request, self.api)
-        self._set_app_name_or_error(data.get('app_name'))
-        self._set_model_name_or_error(data.get('model_name'))
-        self._set_model_id_or_error(data.get('model_id'))
-        if not self.error:
-            self._set_error()
-        if self.error:
-            return get_response_for_bad_request(why=self.error, api=self.api)
+        if not self.has_permission(request):
+            return self.handle_no_permission()
         return super().dispatch(request, *args, **kwargs)
 
 
-class ParentIdMixin:
-    """
-    When using this with DRF intialize `api=True` in the class.
-    Use this mixin only after ContentTypeMixin otherwise `model_id` might not be defined.
-    Used for provinding validation of query parameter `parent_id` in request to match
-    a valid parent comment. Upon successful validation, `parent_id` can be accessed
-    directly as a class attribute.
-    """
-    api = False
-    error = ''
+class ObjectLevelMixin(BaseCommentMixin):
+    @abc.abstractmethod
+    def get_object(self):
+        raise ImproperlyConfigured(
+            _("Your %s class has not defined a get_object() method, which is required.")
+            % self.__class__.__name__
+        )
 
-    def _clean_parent_id(self, val):
-        if not val or val == '0':
-            val = None
-        return val
 
-    def _set_parent_id_or_error(self, val):
-        if val:
-            model_id = self.data.get('model_id')
-            try:
-                if not Comment.objects.filter(id=val, object_id=model_id).exists():
-                    self.error = _(
-                        f'{val} is NOT a valid id for a parent comment or '
-                        'the parent comment does NOT belong to this model object'
-                        )
-                    return
-            except ValueError:
-                self.error = f"the parent id must be an integer, {val} is NOT"
-                return
-        self.parent_id = val
+class CanEditMixin(ObjectLevelMixin, ValidatorMixin, abc.ABC):
+    def has_object_permission(self, request, obj):
+        return request.user == obj.user
 
     def dispatch(self, request, *args, **kwargs):
-        self.data = get_data_for_request(request, self.api)
-        parent_id = self._clean_parent_id(self.data.get('parent_id'))
-        self._set_parent_id_or_error(parent_id)
-        if self.error:
-            return get_response_for_bad_request(why=self.error, api=self.api)
+        obj = self.get_object()
+        if not self.has_object_permission(request, obj):
+            return self.handle_no_permission()
+        return super().dispatch(request, *args, **kwargs)
+
+
+class CanDeleteMixin(ObjectLevelMixin, ValidatorMixin, abc.ABC):
+    def has_object_permission(self, request, obj):
+        return request.user == obj.user or is_comment_admin(request.user) \
+                or (obj.is_flagged and is_comment_moderator(request.user))
+
+    def dispatch(self, request, *args, **kwargs):
+        obj = self.get_object()
+        if not self.has_object_permission(request, obj):
+            return self.handle_no_permission()
+        return super().dispatch(request, *args, **kwargs)
+
+
+class BaseFlagMixin(ObjectLevelMixin, abc.ABC):
+    def dispatch(self, request, *args, **kwargs):
+        if not getattr(settings, 'COMMENT_FLAGS_ALLOWED', 0):
+            return HttpResponseForbidden(_('Flagging system must be enabled'))
+        return super().dispatch(request, *args, **kwargs)
+
+
+class CanSetFlagMixin(BaseFlagMixin, abc.ABC):
+    def has_object_permission(self, request, obj):
+        """user cannot flag their own comment"""
+        return obj.user != request.user
+
+    def dispatch(self, request, *args, **kwargs):
+        obj = self.get_object()
+        if not self.has_object_permission(request, obj):
+            self.handle_no_permission()
+        return super().dispatch(request, *args, **kwargs)
+
+
+class CanEditFlagStateMixin(BaseFlagMixin, abc.ABC):
+    def has_permission(self, request):
+        return is_comment_admin(request.user) or is_comment_moderator(request.user)
+
+    def dispatch(self, request, *args, **kwargs):
+        obj = self.get_object()
+        if not obj.is_flagged:
+            return HttpResponseBadRequest(_('Object must be flagged!'))
+        if not self.has_permission(request):
+            return self.handle_no_permission()
         return super().dispatch(request, *args, **kwargs)
--- a/comment/models/flags.py
+++ b/comment/models/flags.py
@@ -87,7 +87,7 @@ class Flag(models.Model):
         if not allowed_flags:
             return
         self.refresh_from_db()
-        if self.count > allowed_flags and self.state == self.UNFLAGGED:
+        if self.count > allowed_flags and self.state not in [self.RESOLVED, self.REJECTED]:
             self.state = self.FLAGGED
             self.save()
         else:
--- a/comment/tests/test_mixins.py
+++ b/comment/tests/test_mixins.py
@@ -1,6 +1,11 @@
+from unittest.mock import patch
+
+from django.core.exceptions import ImproperlyConfigured
+from django.urls import reverse
 from rest_framework import status
 
-from comment.mixins import AJAXRequiredMixin, ContentTypeMixin, ParentIdMixin
+from comment.conf import settings
+from comment.mixins import AJAXRequiredMixin, BasePermission, ObjectLevelMixin
 from comment.tests.base import BaseCommentMixinTest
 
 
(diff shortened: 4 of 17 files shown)


comment/validators.py
```diff
@@ -0,0 +1,105 @@
...
+        if not self.model_id:
+            self.error = _("model id must be provided")
+            raise CommentErrorValidation(self.error)
+        if not self.app_name:
+            self.error = _('app name must be provided')
+            raise CommentErrorValidation(self.error)
+
+        cause = 'app'
+        try:
+            ContentType.objects.get(app_label=self.app_name)
```
abhiabhi94, line 66: This doesn't always work. This was fixed in #57 maybe you accepted the earlier version during merge conflict resolution.

comment/validators.py
```diff
@@ -0,0 +1,105 @@
...
+        self.model_name = request.GET.get("model_name") or request.POST.get("model_name")
+        self.model_id = request.GET.get("model_id") or request.POST.get("model_id")
+        self.parent_id = request.GET.get("parent_id") or request.POST.get("parent_id")
+
+
+class ContentTypeValidator(BaseValidatorMixin):
+    def validate(self, request):
+        super().validate(request)
+        if not self.model_name:
+            self.error = _("model name must be provided")
```
abhiabhi94, line 55: Instead of calling the` gettext_lazy` at every error, the code could be kept cleaner by directly calling it only once, inside the class `CommentErrorValidation` when adding the `detail` attribute.

Also, shouldn't it's name be something like `CommentBadRequest`?

comment/views/comments.py
```diff
@@ -69,23 +69,21 @@ def form_valid(self, form):
...
+class UpdateComment(CanEditMixin, BaseCommentView):
     comment = None
 
-    def dispatch(self, request, *args, **kwargs):
+    def get_object(self):
         self.comment = get_object_or_404(Comment, pk=self.kwargs.get('pk'))
-        if request.user != self.comment.user:
-            raise PermissionDenied
-        return super().dispatch(request, *args, **kwargs)
+        return self.comment
```
abhiabhi94, line 86: This may be replaced with the [`SingleObjectMixin`](https://docs.djangoproject.com/en/dev/ref/class-based-views/mixins-single-object/#single-object-mixins)

comment/mixins.py
```diff
@@ -1,113 +1,103 @@
...
+        return super().dispatch(request, *args, **kwargs)
+
 
+class BasePermission(AJAXRequiredMixin):
+    def has_permission(self, request):
+        return True
 
-class CommentUpdateMixin(LoginRequiredMixin):
+    def has_object_permission(self, request, obj):
+        return True
```
abhiabhi94, line 25: Inheriting `AJAXRequiredMixin` here means that only AJAX requests will now be allowed even for DRF?

comment/mixins.py
```diff
@@ -1,113 +1,103 @@
...
-            model_class = ct.objects.filter(id=self.model_id)
-            if not model_class.exists() and model_class.count() != 1:
-                self.error = f"{self.model_id} is NOT a valid model id for the model {self.model_name}"
-        except ContentType.DoesNotExist:
-            self.error = f"{self.model_name} is NOT a valid model name"
-        except ValueError:
-            self.error = f"model id must be an integer, {self.model_id} is NOT"
+class CanSetFlagMixin(BaseFlagMixin, abc.ABC):
+    def has_object_permission(self, request, obj):
+        """user cannot flag their own comment"""
```
abhiabhi94, line 83: any particular reason why? This seems a bit opiniated to me.

Suppose I comment something that seems funny to me today but return 2 months later to become more sensitive and find it may be nasty towards a community. I would want to flag my own comment for the moderators to decide if the comment should be flagged or not is what I think.

But this is petty thing, feel free to ignore

comment/tests/test_mixins.py
```diff
@@ -16,131 +21,175 @@ def test_non_ajax_request(self):
...
-        request = self.factory.get(self.get_url(**url_data))
-        response = self.mixin.dispatch(request)
 
-        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
-        self.assertEqual(self.mixin.error, f'{model_name} is NOT a valid model name')
+class CanEditMixinTest(BaseCommentMixinTest):
+    def setUp(self):
+        super().setUp()
+        self.client.force_login(self.user_1)
+        self.comment = self.create_comment(self.post_1, self.user_1)
```
abhiabhi94, line 69: It would be better if we create objects in the `setUpTestData` method. It would make the tests run faster.

comment/tests/test_views.py
```diff
@@ -211,12 +214,12 @@ def test_cannot_delete_comment_by_different_user(self):
...
         # test GET request
-        response = self.client.get(self.get_url(comment.id), data=self.data)
-        self.response_fails_test(response)
+        _response = self.client.get(self.get_url('comment:delete', comment.id), data=self.data)
+        self.response_fails_test(_response)
 
         # test POST request
-        response = self.client.post(self.get_url(comment.id), data=self.data)
-        self.response_fails_test(response)
+        _response = self.client.post(self.get_url('comment:delete', comment.id), data=self.data)
```
abhiabhi94, line 221: Any particular reason for changing `response` to `_response`?

comment/tests/test_mixins.py
```diff
@@ -16,131 +21,175 @@ def test_non_ajax_request(self):
...
+        response = self.client.post(self.url, data=self.data, HTTP_X_REQUESTED_WITH='XMLHttpRequest')
+        self.assertEqual(response.status_code, 403)
+
+    def test_delete_flagged_comment_by_moderator(self):
+        """moderator cannot delete comment unless it's flagged"""
+        self.assertNotEqual(self.comment.user.id, self.moderator.id)
+        self.client.force_login(self.moderator)
+        self.create_flag_instance(self.user_2, self.comment)
+        self.create_flag_instance(self.admin, self.comment)
+        self.create_flag_instance(self.moderator, self.comment)
```
abhiabhi94, line 127: `create_flag_instance` won't work here since it is a class method now. Please use `set_flag`.

comment/validators.py
Radi85: You are right.

comment/validators.py
Radi85: Yes, why not. I got better name.

comment/views/comments.py
Radi85: For now we don't need to bring a cow for glass of milk :D this might be refactor later

comment/mixins.py
Radi85: These mixins are not used in the API, we are using the default permission. I prefer to keep this logic separate to avoid complexity. However, some api views need to be updated to have the same functionality implemented in the GUI version i.e. deleting comment by admin.

comment/mixins.py
Radi85: Yes. This doesn't make sense.

comment/tests/test_mixins.py
Radi85: What do you mean? I actually used it and the comment indeed became flagged. BTW, it took me so long debugging this test case and finally found a silly bug in changing the flag state method https://github.com/Radi85/Comment/pull/67/files#diff-76e68827e2becc6ab78bf6afd4659de9L90
The state was changing to unflagged after flagging a flagged comment 😄

comment/models/flags.py
```diff
@@ -87,7 +87,7 @@ def toggle_flagged_state(self):
         if not allowed_flags:
             return
         self.refresh_from_db()
-        if self.count > allowed_flags and self.state == self.UNFLAGGED:
+        if self.count > allowed_flags and self.state not in [self.RESOLVED, self.REJECTED]:
```
abhiabhi94, line 90: something that has a state rejected shall never be flagged in future now !!!

The earlier version seemed more appropriate to me.

comment/tests/test_mixins.py
abhiabhi94: `create_flag_instance` is a class method?

> The state was changing to unflagged after flagging a flagged comment smile

why would that happen?  if I'm in my most stable state, self.state == self.FLAGGED would not be true !!!

comment/models/flags.py
Radi85: This is how it should be, when the flag state is resolved or rejected, that means the moderator has viewd the comment and took an action, thus the flag state won't be set to flagged anymore.

comment/tests/test_mixins.py
Radi85:
```python
if self.count > allowed_flags and self.state == self.UNFLAGGED:
```
`self.state == self.UNFLAGGED` this will be false when the comment is already flagged, therefore it will hit the else statment and unflag the comment though `self.count = allowed_flags + 2`

abhiabhi94: i am not sure i completely understand currently why an else statement is necessary there. i will re-check when i get up tomorrow.

abhiabhi94: yes, the else part will be useful when people delete their flags.

abhiabhi94: but why is `self.create_flag_instance` working here?

comment/validators.py
```diff
@@ -0,0 +1,101 @@
...
+class CommentBadRequest(APIException):
+    status_code = status.HTTP_400_BAD_REQUEST
+    default_detail = 'Bad Request'
+
+    def __init__(self, detail=None, status_code=None):
+        if status_code:
+            self.status_code = status_code
+        if not detail:
+            detail = self.default_detail
+        self.detail = detail
```
abhiabhi94, line 20: this can be moved to a separate module, maybe `exceptions.py`.

comment/validators.py
```diff
@@ -0,0 +1,101 @@
...
+            model_class = ct.objects.filter(id=self.model_id)
+            if not model_class.exists() and model_class.count() != 1:
+                self.error = f'{self.model_id} is NOT a valid model id for the model {self.model_name}'
+                raise CommentBadRequest(self.error)
+        except ContentType.DoesNotExist:
+            self.error = f'{self.model_name} is NOT a valid model name'
+            raise CommentBadRequest(self.error)
+        except ValueError:
+            self.error = f'model id must be an integer, {self.model_id} is NOT'
+            raise CommentBadRequest(self.error)
```
abhiabhi94, line 79: Too many if else doesn't always give great readability to the code. This was refactored in an earlier commit, maybe you missed it.

```python

    def _set_error_for_model_name(self, val):
        if not val:
            self.error = "model name must be provided"
           raise CommentBadRequest(self.error)

    def _set_error_for_model_id(self, val):
        if not val:
            self.error = "model id must be provided"
            raise CommentBadRequest(self.error)

    def _set_error_for_app_name(self, val):
        if not val:
            self.error = 'app name must be provided'
            raise CommentBadRequest(self.error)

        if not ContentType.objects.filter(app_label=val).exists():
            self.error = f'{val} is NOT a valid app name'
            raise CommentBadRequest(self.error)

    def _set_error(self):
        try:
            model_name = self.model_name.lower()
            ct = ContentType.objects.get(model=model_name).model_class()
            model_class = ct.objects.filter(id=self.model_id)
            if not model_class.exists() and model_class.count() != 1:
                self.error = f"{self.model_id} is NOT a valid model id for the model {self.model_name}"
        except ContentType.DoesNotExist:
            self.error = f"{self.model_name} is NOT a valid model name"
        except ValueError:
            self.error = f"model id must be an integer, {self.model_id} is NOT"

    def validate(self, request):
         super().validate(request)
         self._set_error_for_model_name(self.model_name)
         self._set_error_for_model_id(self.model_id)
         self._set_error_for_app_name(self.app_name)
         self._set_error()
```

comment/validators.py
```diff
@@ -0,0 +1,101 @@
...
+            Comment.objects.get(id=self.parent_id, object_id=self.model_id)
+        except Comment.DoesNotExist:
+            self.error = (
+                f'{self.parent_id} is NOT a valid id for a parent comment or '
+                'the parent comment does NOT belong to the provided model object'
+            )
+            raise CommentBadRequest(self.error)
+        except ValueError:
+            self.error = f'the parent id must be an integer, {self.parent_id} is NOT'
+            raise CommentBadRequest(self.error)
```
abhiabhi94, line 97: Code readability could be increased if we could refactor this to

 ```python

    def _clean_parent_id(self, val):
        if not val or val == '0':
            val = None
        return val

    def _set_error_for_parent_id(self, val):
        if val:
            model_id = self.data.get('model_id')
            try:
                if not Comment.objects.filter(id=val, object_id=model_id).exists():
                    self.error = _(
                        f'{val} is NOT a valid id for a parent comment or '
                        'the parent comment does NOT belong to this model object'
                        )
                    raise CommentBadRequest(self.error)
            except ValueError:
                self.error = f"the parent id must be an integer, {val} is NOT"
                raise CommentBadRequest(self.error)

    def validate(self, request):
         super().validate(request)
         self.parent_id = self._clean_parent_id(self.parent_id)
         self._set_error_for_parent_id( self.parent_id)

```

File: comment/tests/test_mixins.py [hunk 13]
```diff
@@ -16,131 +21,179 @@ def test_non_ajax_request(self):
...
-        parent_id = 100
-        url_data = self.url_data.copy()
-        url_data['parent_id'] = parent_id
-        request = self.factory.get(self.get_url(**url_data))
-        response = self.mixin.dispatch(request)
+class CanSetFlagMixin(BaseCommentMixinTest):
+    def setUp(self):
+        super().setUp()
+        self.client.force_login(self.user_1)
+        self.comment = self.create_comment(self.post_1, self.user_1)
```
abhiabhi94, line 163: Even `create_comment` is a `classmethod` now !

Also consider moving this to `setUpTestData`

comment/tests/test_mixins.py
Radi85: why should not?

comment/validators.py
Radi85: I believe the old one is better. 3 indentation levels make the readability less 😄

comment/tests/test_mixins.py
abhiabhi94: it is a class method, you're calling it by passing an instance.

comment/validators.py
abhiabhi94: for what it's worth, it also violates [single responsibility principle](https://en.m.wikipedia.org/wiki/Single-responsibility_principle)

comment/tests/test_mixins.py
Radi85: As I know, any class property can be accessed by the instance.

comment/validators.py
```diff
@@ -0,0 +1,127 @@
...
+        return comment
+
+    def validate(self, request):
+        super().validate(request)
+        model_id = request.GET.get("model_id") or request.POST.get("model_id")
+        parent_id = request.GET.get("parent_id") or request.POST.get("parent_id")
+        if not parent_id or parent_id == '0':
+            return
+        self.parent_id = self.validate_parent_id(parent_id)
+        self.validate_comment_object(model_id, self.parent_id)
```
abhiabhi94, line 123: code- readability :heart:

comment/exceptions.py
```diff
@@ -0,0 +1,14 @@
...
+class CommentBadRequest(APIException):
+    status_code = HTTP_400_BAD_REQUEST
+    default_detail = 'Bad Request'
+
+    def __init__(self, detail=None, status_code=None):
+        if status_code:
+            self.status_code = status_code
+        if not detail:
+            detail = self.default_detail
+        self.detail = detail
```
abhiabhi94, line 14: I think you probably forgot to return the translated string here.

Radi85: I didn't. the translated string is already used in json response.

abhiabhi94: Could you please point out the line in the code? I didn't seem to find it right now.

abhiabhi94: could point to the line in the code that does it? as of now, i can't seem to find it.

Radi85: https://github.com/Radi85/Comment/pull/67/files/d40fdf2aa1a3a7996ebaea1600d7824d90daf0e6#diff-75aac02b8fa19905987d843be8ad2212R26

abhiabhi94: correct me if i am wrong, but this wouldn't work for the API?

Radi85:
@abhiabhi94 
I think I am going to release v2.0.0 . Would you like to add something before I proceed?

abhiabhi94:
yes there was something, wait let me share the code link

comment/exceptions.py
Radi85: You are right. But I don't think the api should be translated since it will be handled by the developer in the frontend not the end user.

abhiabhi94:
https://github.com/Radi85/Comment/pull/67/files#diff-76e68827e2becc6ab78bf6afd4659de9L59-L61

I don't think needs to be part of the model object. The manager should be a better place for it i think

Radi85:
> https://github.com/Radi85/Comment/pull/67/files#diff-76e68827e2becc6ab78bf6afd4659de9L59-L61
> 
> I don't think needs to be part of the model object. The manager should be a better place for it i think

The link is taking me to JS file, and I don't think this what you want to point to.

comment/exceptions.py
abhiabhi94: yes you're right the frontend developers can translate it on the frontend, but it would be more wise to send it from the backend. They can obviously not use translation if they don't want.

Also, if we go by this principle, we already send translated string responses at the various places in the API.

abhiabhi94:
https://github.com/Radi85/Comment/pull/67/files#diff-76e68827e2becc6ab78bf6afd4659de9L60

soory I selected a bunch of lines earlier, please check this link

comment/exceptions.py
Radi85: I think we should open a new issue to integrate translations and with this we can cover everything. And we might start adding the translation files for some languages.

abhiabhi94:
also please check the contextual or grammatical errors with the documentation

Radi85:
> https://github.com/Radi85/Comment/pull/67/files#diff-76e68827e2becc6ab78bf6afd4659de9L60
> 
> soory I selected a bunch of lines earlier, please check this link

I am still not able to get is. can you just paste the code.

comment/exceptions.py
abhiabhi94: yes, i think most of the backend response is translated if i can recall correctly. only strings inside the template and js files need to be translated.

Radi85:
> also please check the contextual or grammatical errors with the documentation

I don't think that I am going to read all the docs now 😄 any issue with the docs can be easily fixed later

abhiabhi94:
> I am still not able to get is. can you just paste the code.

Yes, because it was hidden.


```
    @property
    def is_flag_enabled(self):
    def is_flag_enabled(self):
        return bool(getattr(settings, 'COMMENT_FLAGS_ALLOWED', 0))
```

abhiabhi94:
> I don't think that I am going to read all the docs now smile any issue with the docs can be easily fixed later

:laughing:  I can understand, but people are only going to use the library only if they can understand the documentation. ;)

Radi85:
> > I don't think that I am going to read all the docs now smile any issue with the docs can be easily fixed later
> 
> 😆 I can understand, but people are only going to use the library only if they can understand the documentation. ;)

I believe the app is well documented so far.

abhiabhi94:
> I don't think needs to be part of the model object. The manager should be a better place for it i think

If you are okay, i will change and make a pull request as soon as i can.

Radi85:
> > I am still not able to get is. can you just paste the code.
> 
> Yes, because it was hidden.
> 
> ```python
>     @property
>     def is_flag_enabled(self):
>         return bool(getattr(settings, 'COMMENT_FLAGS_ALLOWED', 0))
> ```
> 
> I don't think needs to be part of the model object. The manager should be a better place for it i think

You are right, but this property is used in the template as well so it's more convenient to use it this way `comment.is_flag_enabled`

abhiabhi94:
`allowed_flags` is already being passed to the template context if I remember right?

Radi85:
> > I don't think needs to be part of the model object. The manager should be a better place for it i think
> 
> If you are okay, i will change and make a pull request as soon as i can.

This is not something we need to worry about before releasing a new version. We might consider refactoring if i's really necessary but for now I don't see it this way.

Radi85:
> `allowed_flags` is already being passed to the template context if I remember right?

You are right. This can be moved to the manager. But I won't wait for it 😄

Radi85:
I would like you to open a ticket for the translation issue since you were involved in this subject from the begining. This in case you don't mind 😃

PR #78, opened 2020-09-18
fix README links issues

Radi85:


README.rst
```diff
@@ -262,14 +264,14 @@ Login with:
...
 
 
 Contributing
 ============
 
-For contributing, please see the instructions at `Contributing_`
+For contributing, please see the instructions at Contributing_
 
-.. _`Contributing`: './CONTRIBUTING.rst`
+.. _Contributing: https://github.com/Radi85/Comment/blob/master/CONTRIBUTING.rst
```
abhiabhi94, line 277: The relative link would work for both read the doc and github. this will always take them to github.

Radi85: you may be right. But it doesn't currently in the master

abhiabhi94: it was because of syntax errors i made when referencing the link. it should work now if you do so.

Radi85: The full link would be useful when using readme in the docs.

abhiabhi94: > The full link would be useful when using readme in the docs.

I'm not sure i completely understand this logic.

I just though it would be a bit weird to redirect people to github from read the doc when the file is obviously present there. but this shouldn't be too much of an issue. feel free to ignore and resolve this if you want.

PR #85, opened 2020-09-26
feat: implicitly include static files in the template

Radi85:
include_static templatetag is now deprecated.


abhiabhi94:
sorry, i was late to respond. Yes seems logical to me @Radi85

PR #86, opened 2020-09-30
Fix JS

abhiabhi94:
- remove action attribute from flag request.
- also, add semicolons on lines missing them.

Bug fix

- Fix change in title attribute of `svg` element on flagging.

### Before Bug fix:

comment-flag-bug

### After Bug Fix:
comment-flag

### DIFF
--- a/comment/static/js/comment.js
+++ b/comment/static/js/comment.js
@@ -36,7 +36,7 @@ document.addEventListener('DOMContentLoaded', () => {
 
     // show and hide child comments
     let replyLink = replyLinkElement => {
-        getNthParent(replyLinkElement, 4).nextElementSibling.classList.toggle('d-none')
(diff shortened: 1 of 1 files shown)


Radi85:
Hi @abhiabhi94,
I just realized that you were adding simicolons where they should not. simicolon is not needed after if and else statement.
Screenshot 2020-10-03 at 15 44 25
I believe it's a good sign to change your IDE 😄 
I will fix it

abhiabhi94:
> I just realized that you were adding simicolons where they should not. simicolon is not needed after if and else statement.

Yes, you're right. semicolons should only be used `if-else` end in single line. I'm sorry. Thanks for cleaning up for me  :smile: 
> I believe it's a good sign to change your IDE smile

I have had the sign the day I learn I couldn't use VSCode debugger when running django unit tests. It can currently run when using `unittest`, `pytest` or `nose`. It's proabably just the laziness to avoid change

PR #88, opened 2020-10-01
Prepare application for internationalization

### LINKED ISSUE #70: Add translational capabilities to the application.
I think we should open a new issue to integrate translations and with this we can cover everything. And we might start adding the translation files for some languages.

_Originally posted by @Radi85 in https://github.com/Radi85/Comment/pull/67#discussion_r487401673_

abhiabhi94:
- [x] templates
- [x] backend response
- [x] javascipt responses.

Closes #70


abhiabhi94:
Hey @Radi85, I think to check every string is translated, would be difficult personally. I was thinking to write tests to ensure this.

I came across this blog: http://blog.aptivate.org/en/blog/2011/07/26/checking-missing-translations-automatically/

What do you think about it?

Radi85:
> What do you think about it?

That would be great. This will ensure that any new text (string) added to the app is also added to the translation.

abhiabhi94:
Hey @Radi85, I think adding tests for internationalization can be moved to a separate pull request, since it will require a lot of changes to the HTML too.

I was thinking of adding a class `translatable` to the HTML elements containing translated text, so that it can be used by some sort of selector when parsing, to check for translated elements during tests. Although this alone won't be sufficient since there are a couple of cases where we have a`title` attribute that contains translated string. So, it will have to be a bit more flexible.

If you agree to it, feel free to merge this one if you approve of it.

test/settings.py
```diff
@@ -111,3 +111,5 @@
 COMMENT_PER_PAGE = 4
 
 COMMENT_USE_GRAVATAR = True
+
+COMMENT_ALLOW_TRANSLATION = True
```
Radi85, line 115: What the purpose of making settings' variable? Does it harm if the `JavaScriptCatalog` path is always included?

Radi85 (changes requested):

test/settings.py
abhiabhi94: The path can be always included. The script tag to include the `javascript` file inside [`static.html`](https://github.com/Radi85/Comment/pull/88/files#diff-ac8dcc63ef631044a84d92a272230469R4) should only be included when translations are enabled. This will avoid unnecessary overhead. I will make the change.

comment/urls.py
```diff
@@ -15,3 +16,7 @@
     path('<int:pk>/flag/state/change/', ChangeFlagState.as_view(), name='flag-change-state'),
     re_path(r'^confirm/(?P<key>[^/]+)/$', ConfirmComment.as_view(), name='confirm-comment'),
 ]
```
Radi85, line 18:
```suggestion
    path('jsi18n/', JavaScriptCatalog.as_view(), name='javascript-catalog'),
]
```

and please get rid of code block below:
```python
urlpatterns += [
    path('jsi18n/', JavaScriptCatalog.as_view(), name='javascript-catalog'),
]

Radi85 (changes requested):

comment/urls.py
abhiabhi94: since, this is different all the `URL`s that we have written, i have separated to keep it clean.

Radi85: I see it more confusing this way. I would move this path next to other paths.

Radi85 (approved):

PR #94, opened 2020-10-23
Fix related name clashes with django-waffle app

### LINKED ISSUE #40: Related Name clashes
While trying to migrate the database after the installation, I am running into the following issue:

```
ERRORS:
comment.Flag.moderator: (fields.E304) Reverse accessor for 'Flag.moderator' clashes with reverse accessor for 'Flag.users'.
        HINT: Add or change a related_name argument to the definition for 'Flag.moderator' or 'Flag.users'.
waffle.Flag.users: (fields.E304) Reverse accessor for 'Flag.users' clashes with reverse accessor for 'Flag.moderator'.
        HINT: Add or change a related_name argument to the definition for 'Flag.users' or 'Flag.moderator'.
```

Looks like we have a clash of related_names. Is this a problem anyone else is facing? Let me know if you guys have solved it.

abhiabhi94:
- closes #40


comment/models/flags.py
```diff
@@ -27,7 +27,12 @@ class Flag(models.Model):
     comment = models.OneToOneField(Comment, on_delete=models.CASCADE)
     count = models.PositiveIntegerField(default=0)
     state = models.SmallIntegerField(choices=STATES_CHOICES, default=UNFLAGGED)
-    moderator = models.ForeignKey(User, on_delete=models.SET_NULL, null=True, blank=True)
+    moderator = models.ForeignKey(
```
Radi85, line 30: Do we need moderator field for the flag anymore?

Radi85 (changes requested):

comment/models/flags.py
abhiabhi94: yes, it is used to store information when a user toggles the flag state. See the code here : https://github.com/Radi85/Comment/blob/master/comment/models/flags.py#L82

Also, now that you have raised the point, what do you think about adding a soft delete option that preserves the comment when a `comment_moderator` deletes a comment. When `comment_admins` deletes a comment, I think it could be permanently deleted.

Radi85: > Also, now that you have raised the point, what do you think about adding a soft delete option that preserves the comment when a `comment_moderator` deletes a comment. When `comment_admins` deletes a comment, I think it could be permanently deleted.

Agree. That would be nice.

Radi85 (approved):

Issue #99, opened 2020-11-18
Make the moderation features optional


MrCordeiro:
I'm building a project in a context where moderation or flagging isn't needed. However, because of how the templatetags were done, I'm getting a lot of repeated or similar DB hits when rendering the comment list.

For what I could surmise, there is an extra hit for whenever the function `is_comment_moderator(user)` is called.

Maybe we could simplify all those searches with on/off switch in the product settings?

Another option would be to make some ORM optimization.

abhiabhi94:
If I remember correctly, the flag option is disabled by default.

> Another option would be to make some ORM optimization.

Before going to the database, inside the function `is_comment_admin`,  we can check whether flags have been enabled or not. I think this should be good enough to satisfy our use-case.

If you have the time, it would be nice to accept new contributions to the code.

MrCordeiro:
I think you are talking about these functions, right?

```python
def is_comment_admin(user):
    return user.groups.filter(name='comment_admin').exists() or \
           (user.has_perm('comment.delete_flagged_comment') and user.has_perm('comment.delete_comment'))

def is_comment_moderator(user):
    return user.groups.filter(name='comment_moderator').exists() or user.has_perm('comment.delete_flagged_comment')
```

I think it's the ` user.groups.filter(name='xxxx').exists()` the reason why I'm seeing those extra hits. I just forked the repo. Maybe I can get a PR in time for supper 🤞

Radi85:
You are right, these 2 methods are useful only when flagging system is enabled. So I would return false before hitting the db for both of them when the flagging system is disabled.

abhiabhi94:
> You are right, these 2 methods are useful only when flagging system is enabled

`comment_admins` can delete a comment irrespective of the flagging system. 

I am just pasting the relevant portions of the codebase to show this.

https://github.com/radi85/comment/tree/master/comment/templatetags/comment_tags.py#L108
```python
def can_delete_comment(comment, user):
    return is_comment_admin(user) or (comment.is_flagged and is_comment_moderator(user))
```

https://github.com/Radi85/Comment/blob/master/comment/templates/comment/comments/comment_content.html#L27
```jinja
{% if comment.user == user %}
    | {% include 'comment/comments/edit_icon.html' %}
    {% include 'comment/comments/delete_icon.html' %}
{% elif can_delete %}
    | {% include 'comment/comments/delete_icon.html' %}
    ...
```

abhiabhi94:
merged in #101

PR #101, opened 2020-11-18
Make permission queries optional

MrCordeiro:
PR related to issue #99: extra queries hit the database needlessly if the moderation feature is not being used.

### DIFF
--- a/comment/tests/test_api.py
+++ b/comment/tests/test_api.py
@@ -501,9 +501,14 @@ class APICommentDetailForFlagStateChangeTest(APIBaseTest):
 
     @patch.object(settings, 'COMMENT_FLAGS_ALLOWED', 0)
     def test_change_state_when_flagging_is_disabled(self):
+        response = self.client.post(self.get_url(), data=self.data)
+        self.assertEqual(response.status_code, 403)
+
+    @patch.object(settings, 'COMMENT_FLAGS_ALLOWED', 1)
+    def test_change_state_when_comment_is_not_flagged(self):
         comment = self.comment_2
         self.assertFalse(comment.is_flagged)
-        response = self.client.post(self.get_url(), data=self.data)
+        response = self.client.post(self.get_url(comment.id), data=self.data)
         self.assertEqual(response.status_code, 400)
 
     def test_change_state_by_not_permitted_user(self):
--- a/comment/utils.py
+++ b/comment/utils.py
@@ -79,12 +79,20 @@ def has_valid_profile():
 
 
 def is_comment_admin(user):
-    return user.groups.filter(name='comment_admin').exists() or \
-           (user.has_perm('comment.delete_flagged_comment') and user.has_perm('comment.delete_comment'))
+    if settings.COMMENT_FLAGS_ALLOWED:
+        return user.groups.filter(name="comment_admin").exists() or (
+            user.has_perm("comment.delete_flagged_comment")
+            and user.has_perm("comment.delete_comment")
+        )
+    return False
 
 
 def is_comment_moderator(user):
-    return user.groups.filter(name='comment_moderator').exists() or user.has_perm('comment.delete_flagged_comment')
+    if settings.COMMENT_FLAGS_ALLOWED:
+        return user.groups.filter(name="comment_moderator").exists() or user.has_perm(
+            "comment.delete_flagged_comment"
+        )
+    return False
 
 
 def paginate_comments(comments, comments_per_page, current_page):
--- a/docs/source/settings.rst
+++ b/docs/source/settings.rst
@@ -87,7 +87,6 @@ COMMENT_ALLOW_ANONYMOUS
 
 Should the anonymous commenting featured be allowed? Defaults to ``False``.
 
-
 COMMENT_FROM_EMAIL
 ^^^^^^^^^^^^^^^^^^^
 
--- a/comment/signals/post_migrate.py
+++ b/comment/signals/post_migrate.py
@@ -1,28 +1,31 @@
 from django.contrib.auth.models import Permission, Group
 from django.contrib.contenttypes.models import ContentType
 
+from comment.conf import settings
 from comment.models import Comment
 
 
 def create_permission_groups(sender, **kwargs):
-    comment_ct = ContentType.objects.get_for_model(Comment)
-    delete_comment_perm, __ = Permission.objects.get_or_create(
-        codename='delete_comment',
-        name='Can delete comment',
-        content_type=comment_ct
-    )
-    admin_group, __ = Group.objects.get_or_create(name='comment_admin')
-    admin_group.permissions.add(delete_comment_perm)
-    delete_flagged_comment_perm, __ = Permission.objects.get_or_create(
-        codename='delete_flagged_comment',
-        name='Can delete flagged comment',
-        content_type=comment_ct
-    )
-    moderator_group, __ = Group.objects.get_or_create(name='comment_moderator')
-    moderator_group.permissions.add(delete_flagged_comment_perm)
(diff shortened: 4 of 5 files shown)


MrCordeiro:
I decided to add a second setting instead of using the number of flags, as I find both features are potentially independent: moderation could exist even if the user doesn't want to use the flags.

abhiabhi94:
Hey @MrCordeiro, thanks for the contribution.

From the way, the comment application is written as of now, I think this will require change at a few more places.

I will point out the specific instances in the codebase, when I have some more time today.

abhiabhi94:
> I find both features are potentially independent: moderation could exist even if the user doesn't want to use the flags.

In a situation when flagging is enabled and moderation isn't,  I don't understand the purpose of flagging then.


For all to be on the same page, If I understand correctly, by moderation you mean permissions and groups? Correct me if I didn't get you completely.

If that is in the case, then checking this be sufficient:
```python
def is_comment_admin:
    if settings.COMMENT_FLAG_ALLOWED:
        return user.groups.filter(name="comment_admin").exists() or (
            user.has_perm("comment.delete_flagged_comment")
            and user.has_perm("comment.delete_comment")
        )
    return False
```

MrCordeiro:
Sounds good.
You told be there are other instances of the code that would require changing. Have you had the time to list them?

docs/source/settings.rst
```diff
@@ -87,6 +87,10 @@ COMMENT_ALLOW_ANONYMOUS
 
 Should the anonymous commenting featured be allowed? Defaults to ``False``.
 
+COMMENT_ALLOW_MODERATION
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+Should there be user roles for moderation and admin? If not, set this to `False` in order not to query the database unnecessarily. Defaults to ``True``.
```
abhiabhi94, line 93: Please put double back ticks around `False`.
```suggestion
Should there be user roles for moderation and admin? If not, set this to ``False`` in order not to query the database unnecessarily. Defaults to ``True``.
```

comment/conf/defaults.py
```diff
@@ -8,6 +8,7 @@
 COMMENT_USE_GRAVATAR = False
 
 COMMENT_FLAGS_ALLOWED = 0
+COMMENT_ALLOW_MODERATION = True
```
abhiabhi94, line 11: Maybe you can remove this now that an extra setting is no longer required.

comment/utils.py
```diff
@@ -79,12 +79,20 @@ def has_valid_profile():
 
 
 def is_comment_admin(user):
-    return user.groups.filter(name='comment_admin').exists() or \
-           (user.has_perm('comment.delete_flagged_comment') and user.has_perm('comment.delete_comment'))
+    if settings.COMMENT_ALLOW_MODERATION:
```
abhiabhi94, line 82:
```suggestion
    if settings.COMMENT_FLAGS_ALLOWED`:
```

comment/utils.py
```diff
@@ -79,12 +79,20 @@ def has_valid_profile():
...
+        return user.groups.filter(name="comment_admin").exists() or (
+            user.has_perm("comment.delete_flagged_comment")
+            and user.has_perm("comment.delete_comment")
+        )
+    return False
 
 
 def is_comment_moderator(user):
-    return user.groups.filter(name='comment_moderator').exists() or user.has_perm('comment.delete_flagged_comment')
+    if settings.COMMENT_ALLOW_MODERATION:
```
abhiabhi94, line 91:
```suggestion
    if settings.COMMENT_FLAGS_ALLOWED`:
```

abhiabhi94 (changes requested):
In addition to the already made changes, inside the [`post_migrate signals module`](https://github.com/Radi85/Comment/blob/master/comment/signals/post_migrate.py#L8), the `permission_groups` need not be created nor is there any need to running `adjusting_flagged_comment`. If the developer chooses to have moderation at a later stage, they can toggle the setting and run `./migrate comment` again.

The documentation regarding flagging now, maybe needs to be moved to a separate file. It would state the steps to take when enabling flagging. Although you can safely ignore it, I will add it after this is merged.

comment/utils.py
MrCordeiro: I think there's an extra " ` " at the end there, no?

MrCordeiro (commented):
PR updated according to the comments. 
I also modified the `post_migrate` signal with a similar condition check. However, was unable to write an unit test for it. I have never written a migration test 😬, so I would appreciate any input/help 🙏

abhiabhi94:
> Regarding the migrations test

I assume it would be a bit difficult to test this change because the current `BaseMigrationTest` class works well when we have to go in the migration history from one state to the last state(suppose from 5 to 10). As of now, it takes the name of the migration module to go to. In this case, we'll have to revert to the zeroth state because all permissions and stuff is created whenever the migration is run(we migrate before any test is run, see [`tox.ini`](https://github.com/Radi85/Comment/blob/master/tox.ini#L42) for reference). 

One of the alternative that I think of is adding a separate module for testing. As of now, the `example` application and testing both use the same module. The `testing` module will have flagging disabled by default. You can safely ignore this, I will try to see what best we can do to cover this.


To allow your failing test to pass, you will have to change the function `test_change_state_when_flagging_is_disabled`. This test wasn't actually doing what the test name said :(

Also, you'll have to add a test to cover the new changes.

```python
    # fix for the failing test
    @patch.object(settings, 'COMMENT_FLAGS_ALLOWED', False)
    def test_change_state_when_flagging_is_disabled(self):
        response = self.client.post(self.get_url(), data=self.data)
        self.assertEqual(response.status_code, 403)

    # new test
    @patch.object(settings, 'COMMENT_FLAGS_ALLOWED', 1)
    def test_change_state_when_comment_is_not_flagged(self):
        comment = self.comment_2
        self.assertFalse(comment.is_flagged)
        response = self.client.post(self.get_url(comment.id), data=self.data)
        self.assertEqual(response.status_code, 400)
``` 

There is a bit of cleanup required with these `mixins` and `models`, that I forgot. I will do it after this is merged.

abhiabhi94:
Also, when you are done making the changes, can you please squash your changes to one commit. To keep the git history clean, it would be nice to have this change inside one commit.

MrCordeiro:
But isn't the squash applied as an option when merging the PR?

MrCordeiro (commented):
Updated the tests results

abhiabhi94:
> But isn't the squash applied as an option when merging the PR?

Yes, but it will also include all the messages from all of your commits. I will do it if you have an issue with it.

Thanks for the contribution.

comment/tests/test_api.py
```diff
@@ -504,6 +504,13 @@ def test_change_state_when_flagging_is_disabled(self):
         comment = self.comment_2
         self.assertFalse(comment.is_flagged)
         response = self.client.post(self.get_url(), data=self.data)
+        self.assertEqual(response.status_code, 403)
+        
+    @patch.object(settings, 'COMMENT_FLAGS_ALLOWED', 1)
+    def test_change_state_when_comment_is_not_flagged(self):
+        comment = self.comment_2
+        self.assertFalse(comment.is_flagged)
```
abhiabhi94, line 512: These two lines are required inside this function as now. You may want to remove them,

MrCordeiro (commented):
Squashed commits

abhiabhi94:
@MrCordeiro can you please fix the `flake8` errors.

abhiabhi94 (approved):

abhiabhi94:
Thanks !

MrCordeiro:
Sorry, it took so long. I had totally forgotten about this PR until your last comment.
________________________________

Radi85 (approved):

PR #102, opened 2020-11-19
Fix version number according to PEP440

abhiabhi94:
- version number is now a part of the package and can be accessed by `comment.__version__`.
- use this version when adding javascript urls for translation to improve performance. See: https://docs.djangoproject.com/en/dev/topics/i18n/translation/#note-on-performance

### DIFF
--- a/comment/__init__.py
+++ b/comment/__init__.py
@@ -1,1 +1,16 @@
+import os
+
+
+def _get_version():
+    _parent_project_dir = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+    with open(os.path.join(_parent_project_dir, 'VERSION')) as version_file:
+        version = version_file.read().strip()
+
+    if version.lower().startswith('v'):
+        version = version[1:]
+
+    return version
+
+
+__version__ = _get_version()
 default_app_config = 'comment.apps.CommentConfig'
--- a/comment/tests/test_version.py
+++ b/comment/tests/test_version.py
@@ -1,0 +1,15 @@
+from unittest.mock import patch, mock_open
+
+from django.test import TestCase
+
+from comment import _get_version
+
+
+class TestGetVersion(TestCase):
+    def test_when_versions_startswith_v(self):
+        with patch('builtins.open', mock_open(read_data='v2.0.0'), create=True):
+            self.assertEqual('2.0.0', _get_version())
+
+    def test_when_versions_does_not_startwith_v(self):
+        with patch('builtins.open', mock_open(read_data='2.0.0'), create=True):
+            self.assertEqual('2.0.0', _get_version())
--- a/comment/urls.py
+++ b/comment/urls.py
@@ -1,11 +1,14 @@
 from django.urls import path, re_path
+from django.views.decorators.cache import cache_page
 from django.views.i18n import JavaScriptCatalog
 
+from comment import _get_version
 from comment.views import (
     CreateComment, UpdateComment, DeleteComment, SetReaction, SetFlag, ChangeFlagState,
     ConfirmComment)
 
 app_name = 'comment'
+
 
 urlpatterns = [
     path('create/', CreateComment.as_view(), name='create'),
@@ -16,5 +19,9 @@ urlpatterns = [
     path('<int:pk>/flag/state/change/', ChangeFlagState.as_view(), name='flag-change-state'),
     re_path(r'^confirm/(?P<key>[^/]+)/$', ConfirmComment.as_view(), name='confirm-comment'),
     # javascript translations
-    path('jsi18n/', JavaScriptCatalog.as_view(), name='javascript-catalog'),
+    # The value returned by _get_version() must change when translations change.
+    path(
+        'jsi18n/',
+        cache_page(86400, key_prefix='js18n-%s' % _get_version())(JavaScriptCatalog.as_view()),
+        name='javascript-catalog'),
 ]


abhiabhi94:
[PEP440](https://www.python.org/dev/peps/pep-0440/)

[TLDR](https://stackoverflow.com/a/27494005/5317267)(too long didn't read)

comment/__init__.py
```diff
@@ -1 +1,8 @@
 default_app_config = 'comment.apps.CommentConfig'
+
+__version__ = "2.0.0"
```
Radi85, line 3: Why didn't you just read VERSION file here instead of using a static string. I would prefer to change VERSION file on each releasing rather than changing the __init__ file.

comment/__init__.py
```diff
@@ -1 +1,8 @@
...
+
+
+def _get_version():
```
Radi85, line 6: Why do you use `_` prefix in the function name since you are using it outside the module?

docs/source/conf.py
```diff
@@ -19,23 +19,33 @@
...
 import os
+import re
 
 
-def get_version():
-    version_file_dir = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
-    with open(os.path.join(version_file_dir, 'VERSION')) as version_file:
-        ver = version_file.read().strip()
-    return ver
+docs_dir = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
```
Radi85, line 25: To be more specific and explicit 
```suggestion
docs_parent_dir = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
```

docs/source/conf.py
```diff
@@ -19,23 +19,33 @@
 
 # -- Project information -----------------------------------------------------
 import os
+import re
 
 
-def get_version():
```
Radi85, line 24: I would say this method was more cleaner.

Radi85 (changes requested):
I would say removing VERSION file led to write complex code to extract the version number from __init__.py file. I would prefer to stick with the old approach for getting the version from `VERSION` file and just keep the changes that are required for JS i18n and comment.__version__

comment/__init__.py
abhiabhi94: I had not seen the manifest file when writing this. I thought we are not shipping with the package to the end users.

Now that you put it this way, this solution seems really silly :(

I actually looked for this the source code of *pip*.

I don't know why they do it like this? Maybe because they don't want users to download one extra file or something else?

Shall I revert the change?

comment/__init__.py
abhiabhi94: I have read in Python(also in Django) `_` is also used to show an internal function(or class or attribute or simply an object) which might be changed in the future without any notice.

comment/__init__.py
Radi85: Yes please.

comment/__init__.py
Radi85: I am not sure about that but what I know is that in python the underscore before the function or the variable means that this member is private and shouldn't be used outside the scope.

abhiabhi94: To quote the official `django` documentation:

> **APIs marked as internal**
   Certain APIs are explicitly marked as “internal” in a couple of ways:
    Some documentation refers to internals and mentions them as such. If the documentation says that something is internal, we reserve the right to change it.
    Functions, methods, and other objects prefixed by a leading underscore (_). This is the standard Python way of indicating that something is private; if any method starts with a single _, it’s an internal API.


Reference: https://docs.djangoproject.com/en/3.1/misc/api-stability/#apis-marked-as-internal

comment/tests/test_version.py
```diff
@@ -0,0 +1,19 @@
...
+from unittest.mock import patch, mock_open
+
+from django.test import TestCase
+
+from comment import _get_version
+
+
+class TestGetVersion(TestCase):
+    _parent_project_dir = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+    version_file = os.path.join(_parent_project_dir, 'VERSION')
```
Radi85, line 11: Since you are mocking theversion value below why would you read the real value from the file? the variable `version_file` is unused.
In my opinion it would be more realistic if you assert the `_get_version()` with the actual value in `version_file`.

Radi85 (changes requested):

comment/tests/test_version.py
abhiabhi94: No, the actual value will keep changing. So the test will also require a change every time there's a change of version.

Regarding the variables, I was using them for debugging something. Forgot to remove. will remove them if everything else looks good to you.

Radi85: > No, the actual value will keep changing. So the test will also require a change every time there's a change of version.

the value is read from the VERSION file so if you compare `version_file.read().strip()` with `_get_version()` will do the trick and will not require to change anything when the version is incremented.

However, feel free to use either way!

abhiabhi94: > the value is read from the VERSION file so if you compare version_file.read().strip() with _get_version() will do the trick

It will not be in every case. Suppose I mock the contents to be `v2.0.0`. The function `_get_version()` would return `2.0.0` which would not be the same as contents of the file.

Radi85 (approved):

hossein-shahsavan:
hi how can i use persian calender instead of the deafult calender for comments???

hossein-shahsavan:
for changing the calender i must change the comment model but i dont have access

PR #103, opened 2020-11-19
Fix js translation strings so that makemessages detects them

abhiabhi94:
- add a `gettext` function to javascript when translation is disabled.


Radi85:
Well done!
More cleaner work.

PR #113, opened 2020-12-06
fix: rename user and profile serializers

### LINKED ISSUE #98: Conflict with some third party library when using drf-yasg
**Environment**
* python 3.8
* django 3.1
* django-comments-dab 2.0.0

**Describe the bug**
I use django comments library from third party developers in my project and recently decided to add `drf-yasg`. However I get an error after configuring it and opening the endpoint `/swagger':

```
drf_yasg.errors.SwaggerGenerationError: Schema for <class 'djoser.serializers.UserSerializer'> would override distinct serializer <class 'comment.api.serializers.UserSerializer'> because they implicitly share the same ref_name; explicitly set the ref_name attribute on both serializers' Meta classes
```

Is there any way to add this `ref_name` attribute in to userserializer class of `comment` and/or `djoser`?

**To Reproduce**
Steps to reproduce the behavior (if not an installation or configuration issue):
1. Install djoser and configure it
2. install comment and configure it
3. install drf-yasg, configure it and try to open the `swagger` endpoint that is described by drf-yasg library

Radi85:
- closes #98

### DIFF
--- a/comment/api/serializers.py
+++ b/comment/api/serializers.py
@@ -28,13 +28,13 @@ def get_user_fields():
     return 'id', 'username', 'email', 'profile'
 
 
-class ProfileSerializer(serializers.ModelSerializer):
+class ProfileSerializerDAB(serializers.ModelSerializer):
     class Meta:
         model = get_profile_model()
         fields = getattr(settings, 'COMMENT_PROFILE_API_FIELDS', '__all__')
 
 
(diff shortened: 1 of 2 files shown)


abhiabhi94 (commented):
something like `DABUserSerializer` seems better to me but feel free to use this one as well.

PR #121, opened 2020-12-13
test: Use pytest for testing

### LINKED ISSUE #112: Reduce testing time in  travis
### Change Type
[ ] Refactor/Enhancement
[ ] Testing
[x] CI/CD
[ ] Other

### Proposed Changes

Reduce the testing time in travis by reducing the number of python and django compatibility versions.
We can use the most first and most recent version that our app is compatible with.
#### For python:
- 3.6
- 3.8

#### django:
- 2.1
- 3.1
- django-master

### LINKED ISSUE #119: Use pytest-django for testing
### Change Type
[ ] Refactor/Enhancement
[x] Testing
[ ] CI/CD
[ ] Other

### Proposed Changes
- only tests that need database access can be marked explicitly -> this will help in reducing the time required to run newer tests.(all test cases that import django's `TestCase` are implicitly given the database access.)
- all assertions can be made with the simple `assert` statement without having to remember `assertEqual`. `assertDictEqual` etc
- all existing tests will continue to run, but may be migrated if need arises.
- all other performance improvements of `pytest` can be utilized.
- tests can be run with `pytest` command.

abhiabhi94:
- also fix some tests that required patching settings.
- reducing time for tox tests by running it for lesser environments(still cover all versions)
- add support for python3.9

fix #119
fix #112

### DIFF
--- a/comment/tests/test_models.py
+++ b/comment/tests/test_models.py
@@ -40,16 +40,16 @@ class CommentModelTest(BaseCommentManagerTest):
         comment = self.create_anonymous_comment(posted=timezone.now() - timezone.timedelta(days=1))
         self.assertFalse(comment.is_edited)
 
+    @patch.object(settings, 'COMMENT_FLAGS_ALLOWED', 1)
     def test_replies_method(self):
-        self.assertEqual(self.parent_comment_2.replies().count(), 3)
-
-        settings.COMMENT_FLAGS_ALLOWED = 1
+        init_count = self.parent_comment_2.replies().count()
         reply = self.parent_comment_2.replies().first()
         self.create_flag_instance(self.user_1, reply)
         self.create_flag_instance(self.user_2, reply)
-        # default replies hide flagged comment
-        self.assertEqual(self.parent_comment_2.replies().count(), 2)
-        self.assertEqual(self.parent_comment_2.replies(include_flagged=True).count(), 3)
+        with patch.object(settings, 'COMMENT_SHOW_FLAGGED', False):
+            # default replies hide flagged comment
+            self.assertEqual(self.parent_comment_2.replies().count(), init_count - 1)
+        self.assertEqual(self.parent_comment_2.replies(include_flagged=True).count(), init_count)
 
     @patch('comment.models.comments.hasattr')
     def test_is_flagged_property(self, mocked_hasattr):
--- a/comment/tests/test_utils.py
+++ b/comment/tests/test_utils.py
@@ -1,4 +1,3 @@
-from unittest import TestCase
 from unittest.mock import patch
 
 from django.utils import timezone
@@ -337,36 +336,34 @@ class TestProcessAnonymousCommenting(BaseAnonymousCommentTest, BaseCommentUtilsT
         self.assertEqual(EmailInfo.CONFIRMATION_SENT, response)
 
 
-class UtilsTest(TestCase):
+class TestUtils:
     """Test general purpose utilities that aren't necessarily related to a comment"""
-
-    def setUp(self):
-        self.len_id = 6
+    len_id = 6
 
     def test_id_generator_length(self):
-        self.assertEqual(self.len_id, len(id_generator()))
+        assert self.len_id == len(id_generator())
 
     def test_id_generator_generates_different_ids(self):
-        self.assertNotEqual(id_generator(), id_generator())
+        assert id_generator() != id_generator()
 
     def test_id_generator_prefix(self):
         prefix = 'comment'
         output = id_generator(prefix=prefix)
-        self.assertEqual(True, output.startswith(prefix))
-        self.assertEqual(self.len_id + len(prefix), len(output))
+        assert output.startswith(prefix) is True
+        assert self.len_id + len(prefix) == len(output)
 
     def test_id_generator_suffix(self):
         suffix = 'comment'
         output = id_generator(suffix=suffix)
-        self.assertEqual(True, output.endswith(suffix))
-        self.assertEqual(self.len_id + len(suffix), len(output))
+        assert output.endswith(suffix) is True
+        assert self.len_id + len(suffix) == len(output)
 
     def test_id_generator_chars(self):
-        import string   # flake8:no qa
+        import string
         chars = string.ascii_uppercase
         output = id_generator(chars=chars)
-        self.assertEqual(output, output.upper())
+        assert output == output.upper()
 
     def test_id_generator_len(self):
         len_id = 8
-        self.assertEqual(len_id, len(id_generator(len_id=len_id)))
+        assert len_id == len(id_generator(len_id=len_id))
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,12 +5,9 @@ cache: pip
 
 python:
   - 3.6
-  - 3.7
-  - 3.8
+  - 3.9
 env:
   - DJANGO=2.1
-  - DJANGO=2.2
-  - DJANGO=3.0
   - DJANGO=3.1
   - DJANGO=master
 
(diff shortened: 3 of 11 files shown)


comment/tests/test_utils.py
```diff
@@ -337,36 +336,34 @@ def test_for_drf(self):
...
-        self.len_id = 6
+    len_id = 6
 
     def test_id_generator_length(self):
-        self.assertEqual(self.len_id, len(id_generator()))
+        assert self.len_id == len(id_generator())
 
     def test_id_generator_generates_different_ids(self):
-        self.assertNotEqual(id_generator(), id_generator())
+        assert id_generator() != id_generator()
```
Radi85, line 347: can you please change this test as suggested to get rid of SonarLint error
```suggestion
        first_id = id_generator()
        second_id = id_generator()
        assert first_id != second_id
```

comment/tests/test_utils.py
```diff
@@ -337,36 +336,34 @@ def test_for_drf(self):
         self.assertEqual(EmailInfo.CONFIRMATION_SENT, response)
 
 
-class UtilsTest(TestCase):
+class TestUtils:
```
Radi85, line 339: Why `TestCase` is removed from this test only? I would prefer using the `self.assert*` rather than python assert keyword.

Radi85 (changes requested):
In general I have compared current test with pytest and the current test is still faster. So do you think we need pytest anymore?!
Also can you please remove the text related to the icon from your commit? It's nice on github but let's try to not spam the commits/git history with icon text.

comment/tests/test_models.py
```diff
@@ -40,16 +40,16 @@ def test_is_edited_for_anonymous_comment(self):
         comment = self.create_anonymous_comment(posted=timezone.now() - timezone.timedelta(days=1))
         self.assertFalse(comment.is_edited)
 
+    @patch.object(settings, 'COMMENT_FLAGS_ALLOWED', 1)
```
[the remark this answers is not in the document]
abhiabhi94: no, it is not.

comment/tests/test_utils.py
abhiabhi94: > Why TestCase is removed from this test only?

this is a luxury that you can use with `pytest`. You don't need to pass `unittest`'s `TestCase` or even django's `TestCase` when using `pytest-django`, it does it for you. You can read more about it in their documentation https://docs.pytest.org/en/latest/.

There is a bit of magic, it doesn't necessarily use `python's` `assert` statement. It automatically does some magic behind the scenes and does all the comparison for you so that you don't have to remember and write different statements(`self.assertEqual`, `self.assertTrue`) for different use-cases.

> I would prefer using the self.assert* rather than python assert keyword.

If that is the case, there is very little point in switching to the `pytest`. Feel free to close this one then.

abhiabhi94:
> In general I have compared current test with pytest and the current test is still faster. So do you think we need pytest anymore?!

In order to reduce the time, we will need to re-write most of the current tests. I don't think I meant that at any stage.  What I meant was if we need `pytest`'s  functionality to write newer tests, they can be faster and cleaner.

> Also can you please remove the text related to the icon from your commit? It's nice on github but let's try to not spam the commits/git history with icon text.

okay :cry:, although spam is too strong a word. if it were "spam", it probably wouldn't have been used in framework as popular as `fastapi`(https://github.com/tiangolo/fastapi)

comment/tests/test_utils.py
Radi85: I see that the form of assertion `self.assertEqual` and `self.assertTrue`  is more explicit on the type of the value that's why I would prefer it.
Moreover, I wouldn't prefer to stick with one testing framework.

You have already done the work for #112, if you like you can just revert pytest changes and we can still merge this PR.

comment/tests/test_models.py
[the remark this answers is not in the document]
Radi85: you are right!
The variable name points to a boolean value. I think this name would be more descriptive `COMMENT_MAX_FLAGS_NUMBER_ALLOWED `

Radi85:
> okay 😢, although spam maybe is too strong a word. if it were "spam", it probably wouldn't have been used in framework as popular as `fastapi`(https://github.com/tiangolo/fastapi)

I have two points on this topic:

1. If a style or tech is used in another project, that doesn't mean it is a GOOD choice.
2. Using this fancy icon requires special fonts for the terminal, below is the difference between our git tree and fastapi git tree
Fastapi:
```
* 4fdcdf3 (HEAD -> master, origin/master, origin/HEAD) 📝 Update release notes
* e2a6341 📝 Update Uvicorn installation instructions to use uvicorn[standard] (includes uvloop) (#2543)
* 7046d80 📝 Update release notes
* 9f89399 🌐 Add docs lang selector widget (#2542)
* ed0fe9f 📝 Update release notes
* 6e9b771 🐛 Fix docs overrides directory for translations (#2541)
```

Our tree:
```
* 558ec50 (origin/master, origin/HEAD, master) :recycle: refactor: Clean up views
| * e9b43f1 (pytest) :white_check_mark: test: Use pytest for testing
|/  
* 1c2359e (tag: v2.5.1, staging) fix(#114): fix version issue
* 6de6055 fix: rename user and profile serializers
* 7ca4f11 update issue templates
```
I don't like to see such commit `:white_check_mark: test: Use pytest for testing` in the git history
while I don't mind with this `📝 test: Use pytest for testing`.

comment/tests/test_utils.py
abhiabhi94: The current change to `pytest` was made after the issue was discussed. Reverting things because one of us doesn't like the change, seems a lot of waste of time for both of us. This also makes the process more irritating than enjoyable for me.

I think we need to build consensus and be more clear before any code is written so that both of us doesn't waste time.

comment/tests/test_models.py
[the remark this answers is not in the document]
abhiabhi94: > The variable name points to a boolean value.

something like `COMMONT_IS_FLAG_ALLOWED` appears more outrightly boolean to me.

>  I think this name would be more descriptive COMMENT_MAX_FLAGS_NUMBER_ALLOWED

this seems to be more of a whimsical comment to me :laughing: , `COMMENT_MAX_FLAG_ALLOWED` seems better to me but any such change will require a major version bump. Also, if we go down this route, maybe we might need to change almost all of the defaults.

abhiabhi94:
> If a style or tech is used in another project, that doesn't mean it is a GOOD choice.

I never said anything about this being good or bad. You might be going down the route of Strawman argument.


> I don't like to see such commit :white_check_mark: test: Use pytest for testing in the git history
while I don't mind with this 📝 test: Use pytest for testing.

I will do it as you say, but maybe I want more creative freedom around the project. I do agree that the icon text just has somethat of overhead, but I like it :sob:

comment/tests/test_utils.py
Radi85: First of all I would appreciate you contribution to the project.
There should not be an issue if we try something and revert it after finding out that it doesn't fit our need.
In regard to the pytest, I have told you that I never used this framework and didn't have time to fiddle with it before you implement the changes so I am sorry to waste your time.

Lastly, I think it is a good time to create some labels to tag issues that are ready for implementation to avoid wasting people's time :)

abhiabhi94:
@Radi85 it seems wise to me to close this pull-request and make a new one for #112

I'm doing as of what you say, but `pytest` seems a cleaner framework to me to write tests. Maybe you too will realize this someday and remember the day I said this :laughing: 

Also, jokes apart some current tests were actually passing because of the sequence in which `django`'s `unittest` runs them. Independently they should have been patched with mock settings. I will see that I fix that too in the newer one.

Lastly, one more thing, and I kind of feel a bit weird doing this, but wanted to tell you that I was looking for a job. It would be great of you, if you have any opening where you can refer me.

PR #123, opened 2020-12-14
refactor: Clean up views

### LINKED ISSUE #122: Clean up views
### Change Type
[x] Refactor/Enhancement
[ ] Testing
[ ] CI/CD
[ ] Other

### Proposed Changes

Since, `model_obj` is already present inside the `ContentTypeValidator` `mixin`, it makes more sense to pass it directly to the views. 

https://github.com/Radi85/Comment/blob/57a577018c0f2a88593c4ad3229b54a9e9bcc14b/comment/validators.py#L74-L81

The same is valid for the `parent_comment` object and its respective validator mixin.

This will help us remove unnecessary duplication of code in views.

abhiabhi94:
- use logic performed inside validators inside views to remove duplication.

fix #122

### DIFF
--- a/comment/api/serializers.py
+++ b/comment/api/serializers.py
@@ -7,7 +7,7 @@ from rest_framework import serializers
 
 from comment.conf import settings
 from comment.models import Comment, Flag, Reaction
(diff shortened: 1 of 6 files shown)


Radi85 (approved):
nice clean up

PR #126, opened 2020-12-15
get_user_fields from COMMENT_USER_API_FIELDS

### LINKED ISSUE #117: Better UserSerializer
### Environment
* Python version: 3.8

### Proposed Functionality
The get_user_fields function used in UserSerializer has the hardcoded field of username. In some projects we don't use username as the primary field. I had to change the function to make it work with my software.

https://github.com/Radi85/Comment/blob/1c2359e2c11f6efa4d62ba4d260848b3c8e37ead/comment/api/serializers.py#L22-L28

I suggest something like this. 

```python
def get_user_fields():
    user_model = get_user_model()
    fields = user_model._meta.get_fields()
    for field in fields:
        if hasattr(field, "upload_to"):
            return list(set('id', user_model.USERNAME_FIELD, 'email', 'profile', field.name))
    return 'id', 'email', 'profile'
```

### Use Case
API will work with projects that have email based or phone based authentication

### Database Changes
No Database changes required

ZeroCoolHacker:
Fixes #117 
add `COMMENT_USER_API_FIELDS` settings to define UserSerializerDAB fields

### DIFF
--- a/comment/api/serializers.py
+++ b/comment/api/serializers.py
@@ -2,6 +2,7 @@ from django.contrib.auth import get_user_model
 from django.utils import timezone
 
 from django.apps import apps
+from django.models import ImageField
 
 from rest_framework import serializers
 
@@ -22,10 +23,11 @@ def get_profile_model():
 def get_user_fields():
     user_model = get_user_model()
     fields = user_model._meta.get_fields()
+    api_fields = settings.COMMENT_USER_API_FIELDS
     for field in fields:
-        if hasattr(field, "upload_to"):
-            return 'id', 'username', 'email', 'profile', field.name
-    return 'id', 'username', 'email', 'profile'
+        if hasattr(field, "upload_to") and isinstance(field, ImageField):
+            api_fields.append(field.name)
+    return api_fields
 
 
 class ProfileSerializerDAB(serializers.ModelSerializer):
--- a/comment/api/serializers.py
+++ b/comment/api/serializers.py
@@ -2,7 +2,7 @@ from django.contrib.auth import get_user_model
 from django.utils import timezone
 
 from django.apps import apps
-from django.models import ImageField
+from django.db.models import ImageField
 
 from rest_framework import serializers
 
--- a/comment/api/serializers.py
+++ b/comment/api/serializers.py
@@ -23,11 +23,13 @@ def get_profile_model():
 def get_user_fields():
     user_model = get_user_model()
     fields = user_model._meta.get_fields()
-    api_fields = settings.COMMENT_USER_API_FIELDS
+    api_fields = set(settings.COMMENT_USER_API_FIELDS)
+    api_fields.add('profile')
     for field in fields:
         if hasattr(field, "upload_to") and isinstance(field, ImageField):
-            api_fields.append(field.name)
-    return api_fields
+            api_fields.add(field.name)
+    return list(api_fields)
+
 
 
 class ProfileSerializerDAB(serializers.ModelSerializer):
--- a/comment/api/serializers.py
+++ b/comment/api/serializers.py
@@ -32,7 +32,7 @@ def get_user_fields():
     for field in fields:
         if isinstance(field, ImageField) and hasattr(field, "upload_to"):
             api_fields.append(field.name)
-    return list(api_fields)
+    return api_fields
 
 
 class ProfileSerializerDAB(serializers.ModelSerializer):
@@ -173,9 +173,9 @@ class ReactionSerializer(serializers.ModelSerializer):
         users = {'likes': [], 'dislikes': []}
         for instance in obj.reactions.all():
             if instance.reaction_type == instance.ReactionType.LIKE:
-                users['likes'].append({'id': instance.user.id, 'username': instance.user.username})
+                users['likes'].append({'id': instance.user.id, 'username': getattr(instance.user, instance.user.USERNAME_FIELD)})
             else:
-                users['dislikes'].append({'id': instance.user.id, 'username': instance.user.username})
+                users['dislikes'].append({'id': instance.user.id, 'username': getattr(instance.user, instance.user.USERNAME_FIELD)})
         return users
 
 
--- a/comment/api/serializers.py
+++ b/comment/api/serializers.py
@@ -10,6 +10,7 @@ from comment.conf import settings
 from comment.models import Comment, Flag, Reaction
 from comment.utils import process_anonymous_commenting, get_user_for_request, get_profile_instance
 from comment.messages import EmailError
+from collections import OrderedDict
 
 
 def get_profile_model():
@@ -23,13 +24,14 @@ def get_profile_model():
 def get_user_fields():
     user_model = get_user_model()
     fields = user_model._meta.get_fields()
-    api_fields = set(settings.COMMENT_USER_API_FIELDS)
-    api_fields.add('profile')
+    api_fields = settings.COMMENT_USER_API_FIELDS
+    api_fields.append('profile')
+    # remove duplicates while preseving order
+    api_fields = list(OrderedDict((x, 1) for x in api_fields))
     for field in fields:
         if hasattr(field, "upload_to") and isinstance(field, ImageField):
-            api_fields.add(field.name)
+            api_fields.append(field.name)
     return list(api_fields)
-
 
 
 class ProfileSerializerDAB(serializers.ModelSerializer):
--- a/comment/api/serializers.py
+++ b/comment/api/serializers.py
@@ -173,9 +173,14 @@ class ReactionSerializer(serializers.ModelSerializer):
         users = {'likes': [], 'dislikes': []}
         for instance in obj.reactions.all():
             if instance.reaction_type == instance.ReactionType.LIKE:
-                users['likes'].append({'id': instance.user.id, 'username': getattr(instance.user, instance.user.USERNAME_FIELD)})
+                users['likes'].append({
+                    'id': instance.user.id,
+                    'username': getattr(instance.user, instance.user.USERNAME_FIELD)
+                    })
             else:
-                users['dislikes'].append({'id': instance.user.id, 'username': getattr(instance.user, instance.user.USERNAME_FIELD)})
+                users['dislikes'].append({
+                    'id': instance.user.id, 'username': getattr(instance.user, instance.user.USERNAME_FIELD)
+                    })
         return users
 
 
--- a/comment/api/serializers.py
+++ b/comment/api/serializers.py
@@ -1,3 +1,5 @@
+from collections import OrderedDict
+
 from django.contrib.auth import get_user_model
 from django.utils import timezone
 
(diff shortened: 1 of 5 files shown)


comment/models/comments.py
```diff
@@ -69,7 +69,8 @@ def _set_unique_urlhash(self):
 
     def _set_email(self):
         if self.user:
-            self.email = self.user.email
+            if getattr(self.user, 'email', False):
```
abhiabhi94, line 72: You can probably replace this with
```suggestion
            if hasattr(self.user, 'email'):
```

abhiabhi94: also, with this change you will have to change the field `email` on the model to accept `null` values as well. The existing tests might pass because, as of now, you have not written a test to verify this change.

comment/migrations/0009_auto_20200811_1945.py
```diff
@@ -7,8 +7,9 @@
 def set_default_email(apps, schema_editor):
     comment_model = apps.get_model('comment', 'Comment')
     for comment in comment_model.objects.all():
-        comment.email = comment.user.email
-        comment.save(update_fields=['email'])
+        if getattr(comment.user, 'email', False):
```
abhiabhi94, line 10: This can probably be replaced with
```suggestion
        if hasattr(comment.user, 'email'):
```

comment/api/serializers.py
```diff
@@ -2,13 +2,15 @@
...
 from django.apps import apps
+from django.db.models import ImageField
 
 from rest_framework import serializers
 
 from comment.conf import settings
 from comment.models import Comment, Flag, Reaction
 from comment.utils import process_anonymous_commenting, get_user_for_request, get_profile_instance
 from comment.messages import EmailError
+from collections import OrderedDict
```
abhiabhi94, line 13: This line should be at the top. Standard library imports should be placed at the top as per `PEP8`. You may read about the recommendations here: https://www.python.org/dev/peps/pep-0008/#imports

comment/api/serializers.py
```diff
@@ -22,10 +24,14 @@ def get_profile_model():
...
     fields = user_model._meta.get_fields()
+    api_fields = settings.COMMENT_USER_API_FIELDS
+    api_fields.append('profile')
+    # remove duplicates while preseving order
+    api_fields = list(OrderedDict((x, 1) for x in api_fields))
     for field in fields:
-        if hasattr(field, "upload_to"):
-            return 'id', 'username', 'email', 'profile', field.name
-    return 'id', 'username', 'email', 'profile'
+        if hasattr(field, "upload_to") and isinstance(field, ImageField):
```
abhiabhi94, line 32: The first check should be about `ImageField` is what I feel.
```suggestion python
        if isinstance(field, ImageField) and hasattr(field, 'upload_to'):
```

comment/api/serializers.py
```diff
@@ -22,10 +24,14 @@ def get_profile_model():
 def get_user_fields():
     user_model = get_user_model()
     fields = user_model._meta.get_fields()
+    api_fields = settings.COMMENT_USER_API_FIELDS
+    api_fields.append('profile')
```
abhiabhi94, line 28: This just assumes the user has set it to a list. They may also use a tuple. You may probably change this to something like this to be more flexible:
```suggestion python
    api_fields = list(settings.COMMENT_USER_API_FIELDS)
    api_fields.append('profile')
```

comment/api/serializers.py
```diff
@@ -22,10 +24,14 @@ def get_profile_model():
...
+    # remove duplicates while preseving order
+    api_fields = list(OrderedDict((x, 1) for x in api_fields))
```
abhiabhi94, line 30: I may be wrong, but I''m not sure this syntax is correct for `OrderedDict`.

abhiabhi94 (changes requested):
Hey @ZeroCoolHacker, thank you very much for the contribution.

It would be great if you can make some suggested changes. Also, some of your changes need test cases.

In case you need any help, feel free to ask for it.

comment/api/serializers.py
abhiabhi94, line 30: This probably seems to be a little cleaner and lesser complicated to me.
```suggestion
    api_fields = list(OrderedDict.fromkeys(api_fields))
```

comment/models/comments.py
ZeroCoolHacker: What if i add blank email? Blank is set on EmailField
```suggestion
            if getattr(self.user, 'email'):
                self.email = self.user.email
            else:
                self.email = ''
```

abhiabhi94: Sorry, I didn't remember that I had set `null=True` on the `EmailField` for the model. Why would you want to set it again when we have already set it inside the model? I suspect `django` could already do this for you.

Also, in your suggestion I don't quite see the point of prefering `getattr` over `hasattr` where all you want to do is check whether the `user` object has the `email` attribute or not. Am I missing a point here?

ZeroCoolHacker: I am sorry getattr was a typo. I'll push the changes in an hour

abhiabhi94: no issues. there is no need to be sorry, everyone should be allowed to make mistakes.

comment/migrations/0009_auto_20200811_1945.py
```diff
@@ -7,8 +7,9 @@
 def set_default_email(apps, schema_editor):
     comment_model = apps.get_model('comment', 'Comment')
     for comment in comment_model.objects.all():
-        comment.email = comment.user.email
-        comment.save(update_fields=['email'])
+        if hasattr(comment.user, 'email'):
+            comment.email = comment.user.email
+            comment.save(update_fields=['email'])
```
abhiabhi94, line 12: This change will require an additional test case in the [`test_migrations` module](https://github.com/radi85/comment/blob/master/comment/tests/test_migrations.py). The migration test may turn out to be difficult(because of the way they are written and how they run) to deal with at this moment. Feel free to play with them if you want to.

OR 

You can choose to remove this change, I will try to take care of this in the future.

comment/api/serializers.py
```diff
@@ -22,10 +25,14 @@ def get_profile_model():
 def get_user_fields():
     user_model = get_user_model()
     fields = user_model._meta.get_fields()
+    api_fields = list(settings.COMMENT_USER_API_FIELDS)
+    api_fields.append('profile')
```
abhiabhi94, line 29: Why are you appending the `profile` field here when you have already added it to the [defaults](https://github.com/Radi85/Comment/pull/126/files#diff-04e6bdb3bd4b29a0735aa015d5172ee65fa44d84db8c904db1aa157adca2fe46R7)?

As far I can understand, you might not want to change these fields here?

comment/tests/test_api.py
```diff
@@ -659,12 +659,17 @@ def test_get_profile_model(self):
...
+        # change fields
+        patch.object(settings, 'COMMENT_USER_API_FIELDS', ['email', ]).start()
+        fields = get_user_fields()
+        self.assertEqual(fields, ['email', 'profile', ])
 
         mocked_hasattr = patch('comment.api.serializers.hasattr').start()
         mocked_hasattr.return_value = True
         fields = get_user_fields()
-        self.assertEqual(fields, ('id', 'username', 'email', 'profile', 'logentry'))
+        self.assertEqual(fields, ['email', 'profile'])
```
abhiabhi94, line 672: Can you separate this to different functions? It would be a bit easier to read and understand what each test wants to accomplish that way.

abhiabhi94 (changes requested):

comment/api/serializers.py
```diff
@@ -22,10 +25,14 @@ def get_profile_model():
...
+    # remove duplicates while preseving order
+    api_fields = list(OrderedDict.fromkeys(api_fields))
     for field in fields:
-        if hasattr(field, "upload_to"):
-            return 'id', 'username', 'email', 'profile', field.name
-    return 'id', 'username', 'email', 'profile'
+        if isinstance(field, ImageField) and hasattr(field, "upload_to"):
+            api_fields.append(field.name)
+    return list(api_fields)
```
Radi85, line 35: `api_fields ` is already list you don't need to convert it to list again? 
```suggestion
    return api_fields
```

comment/migrations/0009_auto_20200811_1945.py
Radi85: Why do we need to check if the user has an email field? I believe the email field will be always an attribute of the user object since it is already an attribute of django abstract user model. So even if someone subclass django user model and didn't explicitly use the email field, it will still be an attribute of the user object with a blank value.
So I believe this check statement is not needed and it can be removed from all occurrences in the code.

Radi85 (changes requested):
Hey @ZeroCoolHacker,
many thanks for you contribution.
Can you please squash your commits in one?

comment/migrations/0009_auto_20200811_1945.py
abhiabhi94: > I believe the email field will be always an attribute of the user object since it is already an attribute of django abstract user model

No, developers can delete this field using `model._meta` method. Also, one more thing `email` and `username` field shouldn't be hard coded as developers can use their own name for this. We should be using `user_model.USERNAME_FIELD` and `user_model.EMAIL_FIELD` everywhere.

You can see the relevant documentation regarding custom user model here:
https://docs.djangoproject.com/en/dev/topics/auth/customizing/#django.contrib.auth.models.CustomUser

abhiabhi94: That is why I wanted the author to ignore these changes, and I planned to incorporate these in a new one.

Radi85: Nice observation, we need to open a new ticket to adjust all occurrences of `user.email` and `user.username`.

abhiabhi94:
Hey @ZeroCoolHacker, it would be nice if you could make the requested changes.

ZeroCoolHacker:
> Hey @ZeroCoolHacker, it would be nice to have this merged if you could make the requested changes.

I am sorry i have been busy. I'll push them tomorrow

Radi85:
@abhiabhi94 @ZeroCoolHacker 
Can we please finish this PR?
In case that is not possible please notify me so that I can complete it.

abhiabhi94:
> Can we please finish this PR?
In case that is not possible please notify me so that I can complete it.

I will try to finish this on the weekend. I just wanted to give the original author some time since they said they were busy. Since this is actually blocking #128, it would be nice to have this finished as soon as possible.

abhiabhi94:
@ZeroCoolHacker do you plan to complete this? If not please notify so that I can complete the work.

ZeroCoolHacker:
@abhiabhi94 You can complete this. I can not write the migration testcase

abhiabhi94:
> You can complete this. I can not write the migration testcase

@ZeroCoolHacker  You don't have to write a test case for migrations to complete this. You just need to incorporate the changes requested and squash all of your commits into one.

If you feel you need any help there, feel free to ask.

PR #127, opened 2020-12-23
feat(#43): send email notifications

### LINKED ISSUE #43: Notification message via E-Mail

Radi85:
- Add Follower model.
- Parent comment and main thread can now be subscribed.
- Follow and unfollow functionality is integrated in the UI and the API
- Email notification is sent to all thread followers when creating a
  comment in it.
- Tests clean up.
- Adjust translation tests.

closes #43


README.rst
```diff
@@ -53,6 +53,12 @@
 django-comments-dab
 ===================
 
+Thanks https://www.pythonanywhere.com/
+
+**Here is a** live_ **demo**
+
+.. _live: https://rmustafa.pythonanywhere.com/
+
```
abhiabhi94, line 61: :dancers:

comment/api/permissions.py
```diff
@@ -27,7 +27,7 @@ class FlagEnabledPermission(permissions.BasePermission):
     This will check if the COMMENT_FLAGS_ALLOWED is enabled
     """
     def has_permission(self, request, view):
-        return bool(getattr(settings, 'COMMENT_FLAGS_ALLOWED', 0))
+        return getattr(settings, 'COMMENT_FLAGS_ALLOWED', False)
```
abhiabhi94, line 30: I believe this should still be `bool`.

comment/locale/hi/LC_MESSAGES/djangojs.po
```diff
@@ -17,58 +17,63 @@ msgstr ""
...
-#: comment/static/js/comment.js:355
+#: comment/static/js/comment.js:359
 msgid "Reaction couldn't be processed!, please try again"
 msgstr "प्रतिक्रिया संसाधित नहीं की जा सकी!, कृपया पुनः प्रयास करें"
 
-#: comment/static/js/comment.js:462
+#: comment/static/js/comment.js:394
+msgid "Subscription couldn't be processed!, please try again"
+msgstr "सदस्यता संसाधित नहीं हो सकी!, कृपया पुनः प्रयास करें"
+
```
abhiabhi94, line 52: Have you added the translation for these? or is `django` trying to be smart?

comment/locale/hi/LC_MESSAGES/django.po
```diff
@@ -93,7 +96,9 @@ msgstr "{model_id} मॉडल के लिए मान्य मॉडल 
 msgid ""
 "{parent_id} is NOT a valid id for a parent comment or the parent comment "
 "does NOT belong to the provided model object"
-msgstr "{parent_id} मूल टिप्पणी के लिए मान्य आईडी नहीं है या मूल टिप्पणी प्रदान की गई मॉडल ऑब्जेक्ट से संबंधित नहीं है"
+msgstr ""
+"{parent_id} मूल टिप्पणी के लिए मान्य आईडी नहीं है या मूल टिप्पणी प्रदान की गई मॉडल "
+"ऑब्जेक्ट से संबंधित नहीं है"
```
abhiabhi94, line 101: has there been a change around the length after which strings are wrapped in a line? I can't seem to see any other change here.

comment/api/permissions.py
```diff
@@ -36,3 +36,17 @@ def has_permission(self, request, view):
...
+    This will check if the COMMENT_ALLOW_SUBSCRIPTION is enabled
+    """
+    def has_permission(self, request, view):
+        return getattr(settings, 'COMMENT_ALLOW_SUBSCRIPTION', False)
+
+
+class GetSubscribersPermission(permissions.BasePermission):
+    def has_permission(self, request, view):
+        is_system_enabled = getattr(settings, 'COMMENT_ALLOW_SUBSCRIPTION', False)
+        return (is_comment_admin(request.user) or is_comment_moderator(request.user)) and is_system_enabled
```
abhiabhi94, line 52: Shouldn't the first check be for `is_system_enabled`? In the present state, it will make unnecessary queries to the database when this setting in not used.

comment/locale/hi/LC_MESSAGES/djangojs.po
Radi85: I added with google translation 😄 This can be adjusted later in case of bad translation. But this is needed in the development now to make sure that every string tagged for translation is translated.

comment/api/serializers.py
```diff
@@ -97,6 +98,7 @@ class Meta:
 
     def __init__(self, *args, **kwargs):
         user = kwargs['context']['request'].user
+        self.email_service = None
```
abhiabhi94, line 101: I can't seem to find if this initialization is used at a later stage in this method.

comment/conf/defaults.py
```diff
@@ -26,3 +26,5 @@
 COMMENT_USE_EMAIL_FIRST_PART_AS_USERNAME = False
 
 COMMENT_ALLOW_TRANSLATION = False
+
+COMMENT_ALLOW_SUBSCRIPTION = False
```
abhiabhi94, line 30: Shouldn't this feature be enabled by default? I like this option.

comment/api/serializers.py
Radi85: It's used here https://github.com/Radi85/Comment/pull/127/files/e81dd3dd749bb16afccea6f4d139ac039c6622cb#diff-580565eef493475112cb9daed5cc31fd758b926848bc1ba626bab78ea81acc96R130-R134

comment/managers/followers.py
```diff
@@ -0,0 +1,50 @@
...
+from comment.utils import get_username_for_comment
+
+
+class FollowerManager(models.Manager):
+    def is_following(self, email, model_object):
+        content_type = ContentType.objects.get_for_model(model_object)
+        follower_qs = self.filter(email=email, object_id=model_object.id, content_type=content_type)
+        if follower_qs.exists():
+            return True
+        return False
```
abhiabhi94, line 13: I think you can directly return `follower_qs.exists()` here.

comment/conf/defaults.py
Radi85: Hmmmm, No bcz the email configurations is a part of this feature therefore the developer must know what is doing when enable it.

comment/locale/hi/LC_MESSAGES/djangojs.po
abhiabhi94: Oh !!! I don't think this is a great decision especially for languages that both of us might not be aware of. This may cause an issue to newcomers too.

abhiabhi94:
Hey @Radi85 , great work.

I think this is a pretty large change to incorporate in one pull request. It would have been easier to review if may be the refactoring of separate portions were done separately.

Now, that you have already done this, I assume it might be really difficult to separate this, also due to dependency issues.

It might take me a couple of days to look at all the changes independently.

Anyways, awesome work.

comment/api/serializers.py
abhiabhi94: Yes, I saw it. My point was, initializing it to `None` is not used anywhere in this method.

comment/locale/hi/LC_MESSAGES/djangojs.po
Radi85: In fact using a language that has no ascii letter at all is a good choice to make sure that the string is translated.
This is the method that checks for translation:
https://github.com/Radi85/Comment/pull/127/files/e81dd3dd749bb16afccea6f4d139ac039c6622cb#diff-c2eb376f4cca55dbd53c16353ae9670ae28e0a9e50d73b2f5676ad105726fb30R40-R45

comment/conf/defaults.py
abhiabhi94: Okay, so ideally we should add the link to somewhere in the first page of documentation so that users are aware of this configuration. I think a lot of people might want to incorporate this feature if they can set up this properly.

From my personal experience, mostly I only look extensively at any software's documentation only if I have an issue with it, or I want to configure it a certain way.

comment/locale/hi/LC_MESSAGES/djangojs.po
abhiabhi94: no, my point is that because there was only 1 language at present, you could do this manually and on your own. What if the applications grows to more audience?

comment/api/serializers.py
Radi85: IMHO is a good practice to initialize an external service as an instance variable rather than using it as a scope varaible and to make an instance variable you should initialize it in the __init__ method otherwise lintpy will complain when using `self.var` inside a method other than __init__ before initialize it in it.

Radi85:
> I think this is a pretty large change to incorporate in one pull request. It would have been easier to review if may be the refactoring of separate portions were done separately.

You are right. In fact I faced some issues with the tests after finishing the follower implementation so I needed to adjaust the tests along with this feature, that's why is a giant PR.

I have adjusted the code where you have commented and I will push the new changes.

If you would like to have more review, let me know so that I can keep the PR open otherwise I am going to merge it and start with #128 .

abhiabhi94:
> If you would like to have more review, let me know so that I can keep the PR open otherwise I am going to merge it and start with #128 .

Yes, I have only viewed changes in 6 files as of now and there are 73 of them :cold_sweat:

comment/api/serializers.py
abhiabhi94: okay okay I didn't see that light colored line in between that indicates a big chunk of code in between hasn't changed. My mistake, sorry to bother you. I agree with the initialization practice.

comment/locale/hi/LC_MESSAGES/djangojs.po
Radi85: So far we are using locale `hi` as a part of our testing procedure to make sure the all strings are added to .po file and are translated. 
This will not test if the new added string is alos added to .po file of other language. For that we can think about different way to test that every string added to app is also added to .po of other languages.
Since the .po file in `hi` is tested we can make it as a start point for us to compare other files with it.

abhiabhi94: there are a couple of issues that come to my mind if we go on to that route.

- internationalization is part of the development and not translation. this just makes translation too as a part of it.
- this makes the process cumbersome. The author has to add translations for all available languages when a change in string is made to the code. 
- The author obviously has a knowledge of very limited number of languages and google translation or any other service is not really sensitive to the intricacies of a language. In this situation I would prefer to have the translation at an earlier version.

comment/urls.py
```diff
@@ -18,6 +19,7 @@
     path('<int:pk>/flag/', SetFlag.as_view(), name='flag'),
     path('<int:pk>/flag/state/change/', ChangeFlagState.as_view(), name='flag-change-state'),
     re_path(r'^confirm/(?P<key>[^/]+)/$', ConfirmComment.as_view(), name='confirm-comment'),
+    path('toggle-follow/', ToggleFollowView.as_view(), name='toggle-follow'),
```
abhiabhi94, line 22: `toggle-subscription` instead of `toggle-follow` seems more grammatically correct to me.

comment/locale/hi/LC_MESSAGES/djangojs.po
Radi85: Cool.
I will generate .po for locale `test` and activate it in the test and we use it to test the translation, we just make sure that we use none ascii letters in it.
In this case we won't interfer with the language translation but this will make sure the tring is translated.
I have already started this implementation, you can adjust hi language translation after merging this PR and there won't be any issue with test.

comment/managers/followers.py
```diff
@@ -0,0 +1,47 @@
...
+        self.follow(comment.email, username, model_object)
+        if comment.is_parent:
+            self.follow(comment.email, username, comment.content_object)
+
+    def filter_for_model_object(self, model_obj):
+        content_type = ContentType.objects.get_for_model(model_obj)
+        return self.filter(content_type=content_type, object_id=model_obj.id)
+
+    def get_emails_for_model_object(self, model_obj):
+        return list(map(lambda x: x.email, self.filter_for_model_object(model_obj)))
```
abhiabhi94, line 47: I'm not sure this is the most optimized query for the database when using `querysets`. As of now this will result in a lot of queries to the database. You can probably do this in one query with this
```python
return self.filter_for_model_object(model_obj).prefetch_related('user').values_list('user__email', flat=True)
```

abhiabhi94: also, when we use keeping the variable name short to `model_obj`, I think we can keep the same for function names as well. In essence what I'm trying to refer here is changing the function name to `self.filter_for_model_obj`, will maybe help us save some of our time from typing.

comment/managers/followers.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+class FollowerManager(models.Manager):
+    def is_following(self, email, model_object):
+        content_type = ContentType.objects.get_for_model(model_object)
+        return self.filter(email=email, object_id=model_object.id, content_type=content_type).exists()
+
+    def follow(self, email, username, model_object):
+        if not email or self.is_following(email, model_object):
+            return None
```
abhiabhi94, line 14: You sometimes write `return` and sometimes `return None`. I understand both do the same work but I'm just curious if there's a reason or just the time of the day :smile:

comment/managers/followers.py
```diff
@@ -0,0 +1,47 @@
...
+        """This method is used to set the comment's creator as a follower of own comment and the parent thread"""
+        if not comment.email:
+            return
+        model_object = comment
+        if not comment.is_parent:
+            model_object = comment.parent
+        username = get_username_for_comment(comment)
+        self.follow(comment.email, username, model_object)
+        if comment.is_parent:
+            self.follow(comment.email, username, comment.content_object)
```
abhiabhi94, line 40: Too many `if else` here. Maybe there's an opportunity to make this more readable. I will try to understand more once I can clone this on my machine and see if I can suggest a refactor here.

comment/models/followers.py
```diff
@@ -0,0 +1,21 @@
...
+
+from comment.managers.followers import FollowerManager
+
+
+class Follower(models.Model):
+    email = models.EmailField()
+    username = models.CharField(max_length=50)
+    content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE)
+    object_id = models.PositiveIntegerField()
+    content_object = GenericForeignKey()
```
abhiabhi94, line 13: why are you adding a generic foreignkey here? If I understand correctly, this model will only be associated with a comment. Also, the email field can be `None`.

comment/managers/followers.py
Radi85: 👓 I don't see many `ifs` and there is no `else` at all? 😄

comment/models/followers.py
Radi85: Follower can be associated with Comment model and with Content type that uses the comment.

abhiabhi94: okay, got it. if i have two models, i can choose to associate this with only one of them. nice, really nice.

comment/templates/comment/email/footer.html
```diff
@@ -0,0 +1,12 @@
+<table role="presentation" border="0" cellpadding="0" cellspacing="0">
+    <tr>
+        <td class="content-block">
+            <span class="apple-link">Company Inc, 3 Abbey Road, San Francisco CA 94102</span>
```
abhiabhi94, line 4: Do we need to include the company address for here?

Radi85: 😄 I leave it to the developer to change it, even if we use our custom one, the developer will definitely change it to their need so why bother to change it?

comment/managers/followers.py
Radi85: Nice question. In fact I use `return None` when the method explicitly returns a value and just `return` for void methods. I know both of them qualify to `None` but I feel it better for readability to use it this way. So whe I see `return` I know that this method is void in general.

comment/templates/comment/email/footer.html
abhiabhi94: > the developer will definitely change it to their need so why bother to change it?

I do get your point.

They should but that doesn't mean they would :smile: They are people like me who aren't as attentive to these small details as others. In my opinion, it might be better to not include these here as some may forget to change it and see it later.

Radi85: You are right, I will disable the footer and it can be added optionally.
I also was thinking of implementing unsubscribe view but I saw the current PR is already giant.

abhiabhi94: too giant to say the least. please don't :bow: . I already see an unnecessary change log also in there.

abhiabhi94:
Hey @Radi85, I think I'm done for the day, and I have only viewed around 20 which also includes the `html` and `css` templates that I just skipped through.

I think it might take me this complete weekend to completely review all parts.

Radi85:
> Hey @Radi85, I think I'm done for the day, and I have only viewed around 20 files which also includes the `html` and `css` templates that I just skipped through.

Thanks for your time @abhiabhi94 
Take your time I will keep this PR open.

comment/locale/hi/LC_MESSAGES/djangojs.po
Radi85: I will revert the .po changes in hi language to avoid the issue mentioned above. we are using no test langauge instead.

abhiabhi94: no, the changes seem fine to me at a glance. your hindi seems to be great :stuck_out_tongue_closed_eyes:

comment/managers/followers.py
Radi85: Since we are going to use `EMAIL_FIELD` instead of email, I believe we will need the current usage.
```python
return list(map(lambda x: x.EMAIL_FIELD, self.filter_for_model_object(model_obj)))
```
I am not sure how this can be done with string chain `user__email`

abhiabhi94: We can just replace the hard coded `email` string with the one obtained from `model.EMAIL_FIELD`.

```python
from django.contrib.auth import get_user_model

User = get_user_model()
email_field = User.EMAIL_FIELD

emails = self.filter_for_model_object(model_obj).prefetch_related('user').values_list(f'user__{email_field}', flat=True)
```
Is there an issue with this? Even if this didn't work for some reason, there are a couple of other ways to perform this query in a more optimized manner than the one currently written.

Also, the current state is not only the most optimized form, but it doesn't actually return a `queryset` too.

abhiabhi94:
> > Hey @Radi85, I think I'm done for the day, and I have only viewed around 20 files which also includes the `html` and `css` templates that I just skipped through.
> 
> Thanks for your time @abhiabhi94
> Take your time I will keep this PR open.

Pleasure :love_you_gesture:

comment/managers/followers.py
Radi85: Cool. I am not sure if this approach works! but why do you think that your query is more optimised?

abhiabhi94:
Hey @Radi85, I'm sorry I couldn't review the code today. I will try to see how much I can review today and complete the rest tomorrow probably.

I think your last commit aeb89e1096faeb3a8c1326ab4c894f36a80a57c7 doesn't actually contain the changes as the commit message says. Maybe you can either squash the two commits, separate these or do something else to fix this.

comment/managers/followers.py
abhiabhi94: > I am not sure if this approach works!
It works i have tested it on my machine.

> why do you think that your query is more optimised?

This was a mistake actually on my part. I actually thought you were not using `email` field on the model and this query would make a hit to the database for obtaining email from the user table for each value in the list(I know even this presumption was wrong). 

Now, that I see the code with my eyes open and brain working(hopefully), the query seems correct, but you should just use the `values_list` part from my suggestion to make this a `queryset`.

comment/api/permissions.py
```diff
@@ -36,3 +36,18 @@ def has_permission(self, request, view):
...
     def has_object_permission(self, request, view, obj):
         return obj.is_flagged and (is_comment_admin(request.user) or is_comment_moderator(request.user))
+
+
+class SubscriptionEnabledPermission(permissions.BasePermission):
+    """
+    This will check if the COMMENT_ALLOW_SUBSCRIPTION is enabled
+    """
+    def has_permission(self, request, view):
+        return getattr(settings, 'COMMENT_ALLOW_SUBSCRIPTION', False)
```
abhiabhi94, line 46: I also don't quite understand why we use default values in our codebase with `getattr` when we already initialize these attributes in the `defaults.py` file.

comment/api/permissions.py
```diff
@@ -36,3 +36,18 @@ def has_permission(self, request, view):
...
+
+
+class GetSubscribersPermission(permissions.BasePermission):
+    def has_permission(self, request, view):
+        if not getattr(settings, 'COMMENT_ALLOW_SUBSCRIPTION', False):
+            return False
```
abhiabhi94, line 52: I think instead of repeating the code here, we can probably inherit the `SubscriptionEnabledPermission` class here. Also, since all of these classes are already inside the `permissions` module, it makes more sense to me use smaller names like `SubscriptionEnabled` and so on.

comment/api/views.py
```diff
@@ -131,15 +133,46 @@ def post(self, request, *args, **kwargs):
...
+        if temp_comment.why_invalid == CommentFailReason.EXISTS:
             return Response({'detail': EmailError.USED_VERIFICATION_LINK}, status=status.HTTP_200_OK)
 
-        return Response(CommentSerializer(comment.obj).data, status=status.HTTP_201_CREATED)
+        comment = temp_comment.obj
+        comment.save()
+        comment.refresh_from_db()
+        if settings.COMMENT_ALLOW_SUBSCRIPTION:
+            self.email_service = DABEmailService(comment, request)
+            self.email_service.send_notification_to_followers()
```
abhiabhi94, line 153: Maybe instead of doing this here inside the `views`, we can probably register this function to a signal that uses the email service to send messages directly when we parent comments are created.

comment/views/comments.py
```diff
@@ -59,14 +60,18 @@ def form_valid(self, form):
...
         if settings.COMMENT_ALLOW_ANONYMOUS and not user:
             # send response, please verify your email to post this comment.
-            response_msg = process_anonymous_commenting(self.request, _comment)
-            messages.info(self.request, response_msg)
+            self.email_service.send_confirmation_request()
+            messages.info(self.request, EmailInfo.CONFIRMATION_SENT)
         else:
             _comment.save()
+            if settings.COMMENT_ALLOW_SUBSCRIPTION:
+                self.email_service.send_notification_to_followers()
```
abhiabhi94, line 74: same thing as the comment on the `api` `views` file.

comment/views/comments.py
```diff
@@ -116,17 +121,22 @@ def post(self, request, *args, **kwargs):
...
+        if not temp_comment.is_valid:
             return render(request, template_name='comment/anonymous/discarded.html')
 
-        return redirect(comment.obj.get_url(request))
+        comment = temp_comment.obj
+        comment.save()
+        comment.refresh_from_db()
+        if settings.COMMENT_ALLOW_SUBSCRIPTION:
+            self.email_service = DABEmailService(comment, request)
+            self.email_service.send_notification_to_followers()
```
abhiabhi94, line 141: using the signal will also allow us to remove this duplication here as the one used in different `view` functions.

comment/views/comments.py
```diff
@@ -116,17 +121,22 @@ def post(self, request, *args, **kwargs):
...
             messages.warning(request, EmailError.USED_VERIFICATION_LINK)
-
-        if not comment.is_valid:
+        if not temp_comment.is_valid:
             return render(request, template_name='comment/anonymous/discarded.html')
 
-        return redirect(comment.obj.get_url(request))
+        comment = temp_comment.obj
+        comment.save()
+        comment.refresh_from_db()
```
abhiabhi94, line 138: also, this was earlier handled by the `process_anonymous_comment` function to avoid this duplication. The same is used in the corresponding api view. Maybe there's an opportunity for some sort of refactor.

comment/managers/followers.py
```diff
@@ -0,0 +1,48 @@
...
+        return self.filter(email=email, object_id=model_object.id, content_type=content_type).exists()
+
+    def follow(self, email, username, model_object):
+        if not email or self.is_following(email, model_object):
+            return None
+        return self.create(email=email, username=username, content_object=model_object)
+
+    def unfollow(self, email, model_object):
+        content_type = ContentType.objects.get_for_model(model_object)
+        self.filter(email=email, object_id=model_object.id, content_type=content_type).delete()
```
abhiabhi94, line 19: Should be using `get` method instead of `filter` when we expected only one instance here? I think it could be better performance wise.

comment/tests/test_template_tags.py
```diff
@@ -144,6 +144,7 @@ def test_get_username_for_comment(self):
         anonymous_comment = self.create_anonymous_comment()
 
         self.assertEqual(get_username_for_comment(comment), comment.user.username)
+        patch.object(settings, 'COMMENT_USE_EMAIL_FIRST_PART_AS_USERNAME', False).start()
         self.assertEqual(get_username_for_comment(anonymous_comment), settings.COMMENT_ANONYMOUS_USERNAME)
 
         patch.object(settings, 'COMMENT_USE_EMAIL_FIRST_PART_AS_USERNAME', True).start()
```
abhiabhi94, line 150: if different parts of the test use different values for the same attribute maybe when we can split them to separate functions or use a context-manager when patching to make the code more readable.

comment/views/followers.py
```diff
@@ -0,0 +1,61 @@
...
+    def post(self, request, *args, **kwargs):
+        """ Allow authenticated users only, anonymous may be added in the future """
+        email = request.POST.get('email', None)
+        response_class = self.get_response_class()
+        if email:
+            email_validator = EmailValidator(FollowError.EMAIL_INVALID)
+            try:
+                email_validator(email)
+            except ValidationError as error:
+                return response_class({'invalid_email': error.messages}, status=400)
```
abhiabhi94, line 33: this can probably be moved to a some sort of `EmailValidator` mixin. At present, this function seems to contain too much of code to easily understand what is being tried to achieve.

test/settings.py
```diff
@@ -98,11 +98,10 @@
 EMAIL_HOST_USER = os.environ.get('EMAIL_HOST_USER', 'user@domain')
 DEFAULT_FROM_EMAIL = EMAIL_HOST_USER
 EMAIL_HOST_PASSWORD = os.environ.get('EMAIL_HOST_PASSWORD', 'password')
-EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'
+EMAIL_BACKEND = 'django.core.mail.backends.smtp.EmailBackend'
```
abhiabhi94, line 101: `console` backend seems apt for testing to me. Is this really an intended change?

comment/service/email.py
```diff
@@ -0,0 +1,82 @@
...
+        username = get_username_for_comment(self.comment)
+        subject = EmailInfo.NOTIFICATION_SUBJECT.format(username=username, thread_name=thread_name)
+        messages = []
+        text_template = 'comment/notifications/notification.txt'
+        html_template = 'comment/notifications/notification.html'
+        for follower in followers:
+            context = self.get_msg_context(thread_name=thread_name, receiver=follower.username)
+            text_msg, html_msg = self.get_message_templates(text_template, html_template, context)
+            messages.append(self.get_message(subject, text_msg, [follower.email], html_msg=html_msg))
+        self.send_messages(messages)
```
abhiabhi94, line 82: We can probably transform this function to something to along these lines to make this more readable and also separate concerns.

```python
   def get_thread(self):
       if self.comment.is_parent:
           return self.comment.content_object
       return self.comment.parent

   def get_thread_name(self):
       if self.comment.is_parent:
           return str(self.comment.content_object)
       return str(self.comment.parent).split(':')[0]
   
   def get_subject(self, thread_name):
       username = get_username_for_comment(self.comment)
       return EmailInfo.NOTIFICATION_SUBJECT.format(username=username, thread_name=thread_name)
        
   def get_messages(thread_name, receivers):
       text_template = 'comment/notifications/notification.txt'
       html_template = 'comment/notifications/notification.html'
       subject = self.get_subject(thread_name)
       messages = []
       for receiver in receivers:
          context = self.get_msg_context(thread_name=thread_name, receiver=receiver.username)
          text_msg, html_msg = self.get_message_templates(text_template, html_template, context)
          messages.append(self.get_message(subject, text_msg, [receiver.email], html_msg=html_msg))
        return  messages
        
    def send_notification_to_followers(self):
       thread = self.get_thread()
       followers = Follower.objects.filter_for_model_object(thread).exclude(email=self.comment.email)
        if not followers:
            return
        thread_name = self.get_thread_name()
        messages = self.get_messages(thread_name, followers)
        self.send_messages(messages)
```

abhiabhi94:
Hey @Radi85, just wanted to ask whether you use docker for the development of `django` or any other project? If yes, how do you manage autocompletion, debugging etc. for IDE without installing the dependency on local machine?

comment/static/js/comment.js
```diff
@@ -351,11 +361,46 @@ document.addEventListener('DOMContentLoaded', () => {
...
+            else if (data.invalid_email) {
+                form.querySelector('.error').innerHTML = data.invalid_email;
+            }
+            else if (data.following) {
+                followButton.querySelector('.comment-follow-icon').classList.add('user-has-followed');
+                followButton.querySelector('span').setAttribute('title', 'Unfollow this thread');
+                hideModal(followModal);
+            } else {
+                followButton.querySelector('.comment-follow-icon').classList.remove('user-has-followed');
+                followButton.querySelector('span').setAttribute('title', 'Follow this thread');
```
abhiabhi94, line 396: In addition to this change of icons, I think we can also, use the function `createInfoElement` to display a message describing to the user whether they have subscribed/unsubscribed successfully. At the moment, after clicking on the follow icon, i missed the slight change in its color when this request completed.

abhiabhi94 (changes requested):
Hey Radico, I checked your changes locally and wanted to congratulate on the amount of work that you have put forward in this.

In addition to the earlier mentioned review, I just wanted to tell that you probably forgot to register the `Follower` model on the `admin`. This probably all calls for us to maybe add additional tests to see whether all available models have been registered with the `admin` or not. 

You can probably ignore the tests for this pull request. I will add them after this is merged.

comment/conf/defaults.py
Radi85: Let's keep it disabled bu default for now.

comment/api/permissions.py
Radi85: Maybe following the good practice? 😄

abhiabhi94: seems unnecessary duplication of code to me. also, probably defeats the overall purpose of having a `defaults` module.

comment/api/views.py
Radi85: I thought about this in the first place but the issue is that initializing email service needs the request object and it is not available within the signals. I believe there is workaround for this issue but it is not worth for our case.
I do agree with django that the request object shall be used from within the view only.

comment/views/comments.py
Radi85: Note that the email service is not used for notifications only but also used for send confirmation_requests so I don't want to initialize the same service in 2 places.

comment/views/comments.py
Radi85: It's not valid as I commented above.

comment/api/views.py
abhiabhi94: This just means that any possible changes to email service will now have to be done possibly at more than one place because of code duplication in our case. Something to consider maybe.

comment/views/comments.py
Radi85: Please keep the business logic in the right place. Utility function is not the right place to change the database.
Avoiding duplication can be done differently. Maybe by constructing a base class as done for `BaseToggleFollowView`

comment/views/comments.py
abhiabhi94: Not the best of decisions in my view.

comment/tests/test_template_tags.py
Radi85: You are right. We have to use patch decorator for changing settings value in all the test and avoid change settings values inside the test, that will be pain in the as* when a test fails due to a settings value.

abhiabhi94: This did happen to me when I was working on the `pytest` branch, I will open an issue for this separately.

comment/api/permissions.py
Radi85: I am not sure how you get rid of the duplication when you use this `settings.COMMENT_ALLOW_SUBSCRIPTION` instead of this `getattr(settings, 'COMMENT_ALLOW_SUBSCRIPTION', False)` ?
If we will have a comment on each line for our preferences we will not finish this feature in this holiday. 😩

comment/views/followers.py
Radi85: Why would we reinvent the wheel?

test/settings.py
Radi85: Yes. I wanted to test an actual email. I will add it to environment variables for eveyone preference.

comment/service/email.py
Radi85: nice suggestion!

Radi85:
> Hey @Radi85, just wanted to ask whether you use docker for the development of `django` or any other project? If yes, how do you manage autocompletion, debugging etc. for IDE without installing the dependency on local machine?

I don't use docker for development. I tried it before with Pycharm and it will do all the tricks for you. Pycharm has built-ins to ease using docker in development and it was convenient using docker with pycharm.
However, I feel comfortable and have more control over using local setup.

Radi85:
> Hey Radico, I checked your changes locally and wanted to congratulate on the amount of work that you have put forward in this.
> 
> In addition to the earlier mentioned review, I just wanted to tell that you probably forgot to register the `Follower` model on the `admin`. This probably all calls for us to maybe add additional tests to see whether all available models have been registered with the `admin` or not.
> 
> You can probably ignore the tests for this pull request. I will add them after this is merged.

Hey @abhiabhi94,
Thanks a lot for your time! Your review indeed increased the code quality.

I will try to deploy the first beta release after merging this PR. Let me know when you finish your review.

comment/views/followers.py
abhiabhi94: I was suggesting about separating this login to a separate function(refactoring).

abhiabhi94 (approved):
Kudos

abhiabhi94:
Hey @Radi85, I suppose you forgot to add the `.po` files for the `test` language. It might have been created on your machine locally when running `tox` test, or you might have created it manually by running the command `python manage.py compilemessages -l test`. Some tests were failing and this took me a lot of time to figure out. If I had run the tests on my local machine as well, we could have figured this out earlier.

To bypass this sort of condition or having to add additional unnecessary content that may change with time I had mocked the `gettext` functions earlier. Any other ideas to solve this?

Radi85:
Hey @abhiabhi94,
The files are already there https://github.com/Radi85/Comment/pull/127/files#diff-66d5ac964f97c45e0775f9e94dc5d716aefdd048a5fa894f382682666930040c
Otherwise, the test won't pass on the pipeline.

abhiabhi94:
Sorry, replace `.po` with `.mo` files in my above comment. I meant to say the compiled versions of the translations.

abhiabhi94:
> Otherwise, the test won't pass on the pipeline.

The tests pass because inside `tox` file we have the command to compile the translations, `python manage.py compilemessages -l test`.

Radi85:
The compiled files are in .gitignore. https://github.com/Radi85/Comment/blame/develop/.gitignore#L37

abhiabhi94:
Yes, I know this. I'm telling you that I or someone else has to compile the translation files before executing all the tests for a particular version of python. The tests pass when we use `tox` because the command to compile is already added to that file.

Note: as the translations strings change the compiled versions will also keep changing.

abhiabhi94:
To reproduce the issue, you can just delete the `.mo` files for the test language and try to run one of the tests that verify translation or run all tests using `python manage.py test`.

The tests should fail complaining about the absence of translation.

PR #130, opened 2020-12-26
doc(#129): add PR and commit message instructions

### LINKED ISSUE #129: Update CONTRIBUTING.rst for new PR.
### Change Type
[x] Addition
[ ] Correction
[ ] Deprecation
[ ] Cleanup (formatting, typos, etc.)

### Section
[ ] Installation
[ ] Setup
[ ] Usage
[ ] REST API
[ ] Styling
[ ] Settings
[x] Contribution/Development
[ ] Other

### Proposed Changes
Update CONTRIBUTING.rst with the following:
New PR should be merged with the `develop` branch

The commit message should refer to a github open issue if any.

Radi85:
closes #129

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -4,6 +4,39 @@ Contributing
 Django Comments Dab is developed and maintained by developers in an Open Source manner.
 Any support is welcome. You could help by writing documentation, pull-requests, report issues and/or translations.
 
+PR and commit messages
+^^^^^^^^^^^^^^^^^^^^^^^
+
+**Pull Requests**
+
+Refer to `#125`_
+
+.. _`#125`: https://github.com/Radi85/Comment/discussions/125
+
+In order to keep master clean and up to date with the current release, all PR shall be merged with ``develop-RELEASE_MILESTONE`` branch.
+
+``RELEASE_MILESTONE`` should be milestone tag of the issue you work on.
+
+**Commit messages**
+
+Use one commit per issue and try to keep the first line within 50 characters.
+
+Need more? use the commit body.
+
+The commit message should reference the issue number in the header (first line) like so:
+
+feat(*#NUMBER*): YOUR COMMIT
+
+**headers:**
+
+- feat(*#NUMBER*) for adding new feature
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.rst
```diff
@@ -4,6 +4,39 @@ Contributing
...
+
+**headers:**
+
+- feat(*#NUMBER*) for adding new feature
+- fix(*#NUMBER*) for fixing a bug
+- ref(*#NUMBER*) for code refactoring and enhancement
+- test(*#NUMBER*) for adding, fixing or adjusting tests
+- doc(*#NUMBER*) for documentation
+
+The issue number can be skipped if not available..
```
abhiabhi94, line 38: Instead of doing this manually, I would recommend we can probably use to use a `pre-commit` hook of some sort of a tool like `gitlint`. These may be added to the developer dependencies as well.

abhiabhi94: Also, maybe we can use something like this for the commit header to `fix #NUMBER`(*without the brackets*). This will automatically close the issue on successful merging of the request on GitHub. It will essentially help us in reducing the duplication of our work(I don't have to write anything else in the commit message or pull-request text for closing the issue).

But, this would work only for bug-fixes :(.

Radi85: I don't think that we really need a hook for this message. 
I would prefer to keep it with the brackets and explicitly mention the issue in the message body. Sometimes one commit closes 2 issues.

abhiabhi94: > I don't think that we really need a hook for this message.

See, if the hook does the work for you, you don't have to tell other people to conform to these guidelines. Also, from my personal experience, the lesser human attention required to these nitty-gritty things, the better it is to focus on potentially more important things.

Radi85: Every project has a guidlines and someone has to follow them if they like to contribute. I don't see any nitty-gritty things here! and I am not sure why you see an issue from these things?

However, feel free to implement git hook to add this pre-headers to the commit message, personally, I am not sure how to find out what the developer is intended to commit, whether fix, feat or so on.

PR #134, opened 2021-01-03
test(#132): Patch settings object for individual tests

### LINKED ISSUE #132: Patch changes to settings for individual tests
### Change Type
[ ] Refactor/Enhancement
[x] Testing
[ ] CI/CD
[ ] Other

### Proposed Changes

Use the `patch` function as a decorator or context manager depending upon the use case.

abhiabhi94:
- some refactoring of the original code and tests that may aid in hopefully better debugging.
- fixes #132

### DIFF
--- a/comment/tests/test_api/test_serializers.py
+++ b/comment/tests/test_api/test_serializers.py
@@ -27,23 +27,23 @@ class APICommentSerializers(APIBaseTest):
 
     def test_get_profile_model(self):
         # missing settings attrs
-        patch.object(settings, 'PROFILE_APP_NAME', None).start()
-        profile = get_profile_model()
-        self.assertIsNone(profile)
+        with patch.object(settings, 'PROFILE_APP_NAME', None):
+            profile = get_profile_model()
+            self.assertIsNone(profile)
 
         # providing wrong attribute value, an exception is raised
-        patch.object(settings, 'PROFILE_APP_NAME', 'wrong').start()
-        self.assertRaises(LookupError, get_profile_model)
+        with patch.object(settings, 'PROFILE_APP_NAME', 'wrong'):
+            self.assertRaises(LookupError, get_profile_model)
 
         # attribute value is None
-        patch.object(settings, 'PROFILE_APP_NAME', None).start()
-        profile = get_profile_model()
-        self.assertIsNone(profile)
+        with patch.object(settings, 'PROFILE_APP_NAME', None):
+            profile = get_profile_model()
+            self.assertIsNone(profile)
 
         # success
-        patch.object(settings, 'PROFILE_APP_NAME', 'user_profile').start()
-        profile = get_profile_model()
-        self.assertIsNotNone(profile)
+        with patch.object(settings, 'PROFILE_APP_NAME', 'user_profile'):
+            profile = get_profile_model()
+            self.assertIsNotNone(profile)
 
     def test_get_user_fields(self):
         fields = get_user_fields()
@@ -56,19 +56,19 @@ class APICommentSerializers(APIBaseTest):
 
     def test_user_serializer(self):
         # PROFILE_MODEL_NAME not provided
-        patch.object(settings, 'PROFILE_MODEL_NAME', None).start()
-        profile = UserSerializerDAB.get_profile(self.user_1)
-        self.assertIsNone(profile)
+        with patch.object(settings, 'PROFILE_MODEL_NAME', None):
+            profile = UserSerializerDAB.get_profile(self.user_1)
+            self.assertIsNone(profile)
 
         # PROFILE_MODEL_NAME is wrong
-        patch.object(settings, 'PROFILE_MODEL_NAME', 'wrong').start()
-        profile = UserSerializerDAB.get_profile(self.user_1)
-        self.assertIsNone(profile)
+        with patch.object(settings, 'PROFILE_MODEL_NAME', 'wrong'):
+            profile = UserSerializerDAB.get_profile(self.user_1)
+            self.assertIsNone(profile)
 
         # success
-        patch.object(settings, 'PROFILE_MODEL_NAME', 'userprofile').start()
-        profile = UserSerializerDAB.get_profile(self.user_1)
-        self.assertIsNotNone(profile)
+        with patch.object(settings, 'PROFILE_MODEL_NAME', 'userprofile'):
+            profile = UserSerializerDAB.get_profile(self.user_1)
+            self.assertIsNotNone(profile)
 
     @patch.object(settings, 'COMMENT_ALLOW_SUBSCRIPTION', False)
     @patch.object(settings, 'COMMENT_ALLOW_ANONYMOUS', False)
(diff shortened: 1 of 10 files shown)


abhiabhi94:
Hey @Radi85, I just have a suggestion:

we shouldn't be using `assertTrue(val)`, if we want to test whether an object has value `True` or not. `assertTrue` checks whether the expression is `Truthy` or not. It is advisable to use `assertIs(True, val)` in such circumstances.

Reference: https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertFalse

Radi85:
> Hey @Radi85, I just have a suggestion:
> 
> we shouldn't be using `assertTrue(val)`, if we want to test whether an object has value `True` or not. `assertTrue` checks whether the expression is `Truthy` or not. It is advisable to use `assertIs(True, val)` in such circumstances.
> 
> Reference: https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertFalse

All the assertion checks currently used in the tests is valid. We are using assertTrue and assertFalse with boolean values only and this is still valid since assertTrue(val) is equivalent to `bool(expr) is True`whereas `bool(False) is False` and `bool(True) is True`. We are not using this assertion to check if a type exists, it will give wrong result if it used like this: `assertFalse('False')`

abhiabhi94:
> All the assertion checks currently used in the tests is valid. We are using assertTrue and assertFalse with boolean values only and this is still valid since assertTrue(val) is equivalent to bool(expr) is Truewhereas bool(False) is False and bool(True) is True. 

I do remember that this had happened once, One of our tests was passing because an object satisfied `Truthy` expression but wasn't exactly True. I can't exactly the instance where this happened, but it would have been wise to have pointed this, then. It was probably with `COMMENT_FLAGS_ALLOWED`.

Radi85:
> I do remember that this had happened once, One of our tests was passing because an object satisfied `Truthy` expression but wasn't exactly True. I can't exactly the instance where this happened, but it would have been wise to have pointed this, then. It was probably with `COMMENT_FLAGS_ALLOWED`.

`COMMENT_FLAGS_ALLOWED` is an integer, the name is confusing though. However there must not be an issue bcz `bool(0) is False`

comment/tests/test_api/test_serializers.py
```diff
@@ -27,23 +27,23 @@ def comment_count_test(self):
 
     def test_get_profile_model(self):
         # missing settings attrs
-        patch.object(settings, 'PROFILE_APP_NAME', None).start()
-        profile = get_profile_model()
-        self.assertIsNone(profile)
+        with patch.object(settings, 'PROFILE_APP_NAME', None):
```
Radi85, line 30: Why don't we use decorator? We can split the test if it uses patched value more than once.

comment/tests/test_api/test_serializers.py
```diff
@@ -56,19 +56,19 @@ def test_get_user_fields(self):
 
     def test_user_serializer(self):
         # PROFILE_MODEL_NAME not provided
-        patch.object(settings, 'PROFILE_MODEL_NAME', None).start()
-        profile = UserSerializerDAB.get_profile(self.user_1)
-        self.assertIsNone(profile)
+        with patch.object(settings, 'PROFILE_MODEL_NAME', None):
```
Radi85, line 59: Same above!

comment/tests/test_api/test_views.py
```diff
@@ -511,22 +512,29 @@ def test_flag_to_previous_comments(self):
...
         # test is_flagged property
         comment = self.create_comment(self.content_object_1)
         data = self.flag_data
         url = self.get_url(comment.id)
-        settings.COMMENT_FLAGS_ALLOWED = 1
         response = self.client.post(url, data=data)
+
         self.assertEqual(response.status_code, 200)
-        self.assertFalse(response.json()['is_flagged'])
+        self.assertIs(False, response.json()['is_flagged'])
```
Radi85, line 524: As previously discussed, `is_flagged` is a boolean value and assertFalse can be used directly.

Radi85 (changes requested):
There is 2 major points:

- Use decorator for patching settings object and the test uses the settings value more than once can be split for more simplicity.
- Use assertFalse and assertTrue with boolean values.

comment/tests/test_api/test_serializers.py
abhiabhi94: I didn't want to go that route otherwise it would refactoring a lot of tests. Some of them are written as functions as of now, which can be changed to functions and then, use decorators. I believe this can be done in a separate commit.

comment/tests/test_api/test_views.py
abhiabhi94: `assertIs` as mentioned in the documentation is more powerful for this type of test. It also provides a more helpful message when debugging.

abhiabhi94:
```sh
 $ python                                   
Python 3.7.5 (default, Nov  7 2019, 10:50:52) 
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import unittest
>>> test = unittest.TestCase()
>>> test.assertIs.__doc__
'Just like self.assertTrue(a is b), but with a nicer default message.'
```

comment/tests/test_api/test_serializers.py
abhiabhi94: as of now, I have only refactored the ones that I had to debug for this issue.

comment/tests/test_api/test_views.py
Radi85: > `assertIs` as mentioned in the documentation is more powerful for this type of test. It also provides a more helpful message when debugging.

In fact the debug message for assertTrue is pretty much descriptive: `False is not True` do we really need more info about this? 😄

Radi85: As I know, you are a big fan of less typing 🙄

comment/tests/test_api/test_serializers.py
Radi85: I don't mind to keep it this way, but since we are in the subject, why don't we finish it in one time?
If you feel that you are busy, I can continue your wrok.

comment/tests/test_api/test_views.py
abhiabhi94: I peeked into the official django docs, and found that they also prefer to use `assertIs` over `assertTrue`.

> As I know, you are a big fan of less typing roll_eyes

I don't want to get distracted by this candy :laughing:

comment/tests/test_api/test_serializers.py
abhiabhi94: If you feel that you are busy, I can continue your wrok.

No, it is not due to time constraint. I just don't like too big changes in one commit unless it is really necessary. Too many changes in one go, make it harder to review and also litter the `git` history in my opinion. 

Some tests really require refactoring as they test more than one unit at present. I would probably want to do this in a separate commit.

Radi85 (approved):

comment/tests/test_api/test_serializers.py
Radi85: No problem. I will merge this and will open a new issue for splitting unit tests in case I will have time later.

PR #141, opened 2021-01-05
fix(#139): fix url resolving issue

### LINKED ISSUE #139: Unresloved toggle-subscription url
### Environment
* django-comments-dab: 2.6b1

### Describe the bug
Runtime error: django cannot relove toggle-subscription url.

### Steps To Reproduce
1. Install the app version 2.6b1
2. Add settings var `COMMENT_ALLOW_SUBSCRIPTION = True`
3. Run the app.

### Expected behavior
The error should not be raised and the url should be resolved.

### Fix suggestion
Prepend the follow url name in the templates with the `app_name`. This should be done for all `url 'toggle-subscription'` occurrences

E.g.
```jinja
{% url 'comment:toggle-subscription' as follow_url %}
{# url 'toggle-subscription' as follow_url #}
```

Radi85:
fixes #139

### DIFF
--- a/comment/templates/comment/follow/follow.html
+++ b/comment/templates/comment/follow/follow.html
@@ -1,6 +1,6 @@
 {% load i18n %}
 {% load comment_tags %}
-{% url 'toggle-subscription' as follow_url %}
+{% url 'comment:toggle-subscription' as follow_url %}
 {% if user.is_authenticated %}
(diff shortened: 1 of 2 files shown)


abhiabhi94 (approved):

abhiabhi94:
I can't think of any way as of now, but it would be nice if we can detect these before merging.

Radi85:
> I can't think of any way as of now, but it would be nice if we can detect these before merging.

This can be in a seperate ticket. This fix is more important for now.

PR #146, opened 2021-01-10
ref(#144): Move common logic from views to mixin

### LINKED ISSUE #144: Get rid of duplication of code in views and serializers
### Change Type
[ x] Refactor/Enhancement
[ ] Testing
[ ] CI/CD
[ ] Other

### Proposed Changes

Move logic from `view/serializer` for the sending of emails through services to one of the following:
 
-  register  a custom signal inside `post_save` module maybe. This signal can be connected to a receiver function from the `view/serializer` to handle the logic. 
-  make a utility function to do this. 

The same applies for both sending confirmation for anonymous comment and  notification to followers of a comment. 

Django although recommends using the latter as per our case. To quote:

> When to use custom signals?
Signals are implicit function calls which make debugging harder. If the sender and receiver of your custom signal are both within your project, you’re better off using an explicit function call.

Reference: https://docs.djangoproject.com/en/dev/topics/signals/#defining-and-sending-signals

abhiabhi94:
- prevents duplication.
- also rename `thread` attribute in DABEmailService to `_email_thread` -> avoids confusion between comment thread and processing threads.

resolves #144

### DIFF
--- a/comment/api/serializers.py
+++ b/comment/api/serializers.py
@@ -9,7 +9,7 @@ from comment.conf import settings
 from comment.models import Comment, Flag, Reaction
 from comment.utils import get_user_for_request, get_profile_instance
 from comment.messages import EmailError
-from comment.service.email import DABEmailService
+from comment.mixins import CommentCreateMixin
 
 
 def get_profile_model():
@@ -91,7 +91,7 @@ class BaseCommentSerializer(serializers.ModelSerializer):
         return ReactionSerializer(obj.reaction).data
 
 
-class CommentCreateSerializer(BaseCommentSerializer):
+class CommentCreateSerializer(BaseCommentSerializer, CommentCreateMixin):
     class Meta:
         model = Comment
         fields = ('id', 'user', 'email', 'content', 'parent', 'posted', 'edited', 'reply_count', 'replies', 'urlhash')
@@ -117,7 +117,7 @@ class CommentCreateSerializer(BaseCommentSerializer):
         email = validated_data.get('email')
         time_posted = timezone.now()
 
-        comment = Comment(
+        temp_comment = Comment(
             content_object=self.context['model_obj'],
             content=content,
             user=user,
@@ -125,14 +125,7 @@ class CommentCreateSerializer(BaseCommentSerializer):
             email=email,
             posted=time_posted
         )
-        self.email_service = DABEmailService(comment, request)
-        if settings.COMMENT_ALLOW_ANONYMOUS and not user:
-            self.email_service.send_confirmation_request(api=True)
-        else:
-            comment.save()
-            if settings.COMMENT_ALLOW_SUBSCRIPTION:
-                self.email_service.send_notification_to_followers()
-        return comment
+        return self.perform_create(temp_comment, request, api=True)
 
 
 class CommentSerializer(BaseCommentSerializer):
--- a/comment/mixins.py
+++ b/comment/mixins.py
@@ -1,6 +1,7 @@
 import abc
 
 from django.contrib.auth.mixins import LoginRequiredMixin, AccessMixin
+from django.contrib import messages
 from django.core.exceptions import ImproperlyConfigured
 from django.http import HttpResponseBadRequest, HttpResponseForbidden
 
@@ -8,6 +9,8 @@ from comment.conf import settings
 from comment.utils import is_comment_admin, is_comment_moderator
 from comment.validators import ValidatorMixin
 from comment.messages import ErrorMessage, FlagError, FollowError
+from comment.service.email import DABEmailService
+from comment.messages import EmailInfo
 
 
 class AJAXRequiredMixin:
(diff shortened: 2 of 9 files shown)


comment/views/comments.py
```diff
@@ -120,23 +107,25 @@ def post(self, request, *args, **kwargs):
         return render(request, 'comment/comments/base.html', context)
 
 
-class ConfirmComment(View):
+class ConfirmComment(View, BaseCommentCreate):
     email_service = None
 
+    def _handle_invalid_comment(self, comment, request):
```
Radi85, line 113: Please make this method static.

Radi85 (changes requested):

comment/mixins.py
```diff
@@ -29,6 +32,36 @@ class BaseCommentMixin(LoginRequiredMixin, BasePermission):
     pass
 
 
+class BaseCommentCreate:
```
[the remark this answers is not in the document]
abhiabhi94: `CreateComment` seems enough, all of these are inside the `mixin` module so they should be considered as `mixin` I think

comment/views/comments.py
abhiabhi94: how would it use the email service then?

comment/api/serializers.py
```diff
@@ -117,21 +117,15 @@ def create(self, validated_data):
         email = validated_data.get('email')
         time_posted = timezone.now()
 
-        comment = Comment(
+        temp_comment = Comment(
```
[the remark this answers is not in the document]
abhiabhi94: i would still prefer to keep the name `temp_comment` just to emphasize the fact that this comment is still temporary.

comment/api/serializers.py
```diff
@@ -117,21 +117,15 @@ def create(self, validated_data):
...
         )
-        self.email_service = DABEmailService(comment, request)
-        if settings.COMMENT_ALLOW_ANONYMOUS and not user:
-            self.email_service.send_confirmation_request(api=True)
-        else:
-            comment.save()
-            if settings.COMMENT_ALLOW_SUBSCRIPTION:
-                self.email_service.send_notification_to_followers()
+        comment = self.perform_create(temp_comment, request, api=True)
         return comment
```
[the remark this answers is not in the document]
abhiabhi94: I just like it in this form because it tells me explicitly what the function is returning rather than go(while debugging) and see inside the function to see.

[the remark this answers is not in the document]
Radi85: The variable is redundant here. you can add this line while debugging if this the case.

comment/mixins.py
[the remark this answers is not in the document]
Radi85: Yes but you need to go to the import on the top to know that it is imported from mixin, I see django is still using this name convention.

comment/views/comments.py
Radi85: The email service is not used inside this method!!

Radi85: I aksed you to make it static bcz `self` is not used at all.

abhiabhi94: sorry, got confused with another comment.

comment/mixins.py
```diff
@@ -29,6 +32,39 @@ class BaseCommentMixin(LoginRequiredMixin, BasePermission):
...
 
+class CommentCreateMixin:
+    email_service = None
+
+    def _initialize_email_service(self, comment, request):
+        self.email_service = DABEmailService(comment, request)
+
+    def send_notification_to_followers(self, comment, request):
+        if settings.COMMENT_ALLOW_SUBSCRIPTION:
+            self._initialize_email_service(comment, request)
```
Radi85, line 43: This didn't solve the issue, calling the method here is still dplicating the code and creating 2 email service instances. `initialize_email_service` should be called **once** inside the view that inherits this mixin.

Radi85 (changes requested):

comment/mixins.py
Radi85: I see that `send_notification_to_followers` is always called after `perform_save` so it can be called inside `perform_save`.

Radi85 (changes requested):

comment/mixins.py
abhiabhi94: > This didn't solve the issue, calling the method here is still dplicating the code and creating 2 email service instances. initialize_email_service should be called once inside the view that inherits this mixin.

```python
if settings.COMMENT_ALLOW_ANONYMOUS and not comment.user:
        self._handle_anonymous(comment, request, api)
else:
       self.perform_save(comment)
       self.send_notification_to_followers(comment, request)
return comment
```

Only one of the two branches can be executed at a time. I don't see how 2 instances of the email service are created.


>

Radi85: yes you are right. However moving `send_notification_to_followers` to `perform_save` is more cleaner.

abhiabhi94: > I see that send_notification_to_followers is always called after perform_save so it can be called inside perform_save.

nice observation

Radi85 (approved):

PR #151, opened 2021-01-17
ref(#150): unify django views responses

### LINKED ISSUE #149: JS error when replying to a comment anonymously.
### Environment
* django-comments-dab: 2.5.1

### Describe the bug
Creating a child comment by anonymous is no handled in JS.
This alert popup `Unable to post your comment!, please try again` when posting a new child comment.

### Steps To Reproduce
1. Reply to a comment anonymously.

### Expected behavior
A message should appear to inform the user that an email is sent to confirm their comment.

### LINKED ISSUE #150: Unify django views response.
### Change Type
[x] Refactor/Enhancement
[ ] Testing
[ ] CI/CD
[ ] Other

### Proposed Changes

Use JsonResponse for all django views instead of using mix of responses specially the JS Fetch API is used to consume django views.

This might help us in the future to unify django and API views for more cleaner and less bugs code.

Radi85:
resolves #150
fixes #149

- Replace all responses in django views with UTF8JsonResponse.
- Handle anonymouse comment form.
- Refactore get context - remove get_comment_context_data and use DABContext class instead.


comment/validators.py
```diff
@@ -121,3 +122,16 @@ def validate(self, request):
...
+    def __init__(self, email):
+        super().__init__(EmailError.EMAIL_INVALID)
+        self.email = email
+
+    def is_valid(self):
+        try:
+            self(self.email)
+            return True
+        except ValidationError:
+            return False
```
abhiabhi94, line 137: why are we not using the default [`EmailValidator`](https://docs.djangoproject.com/en/dev/ref/validators/#emailvalidator) when we aren't performing any custom validation? The error message can be overriden if required through `EmailValidator(message=my_message)`. 

Although in our case, we don't necessarily need to do this as the default error message is sufficient `Enter a valid email address`.

comment/templates/comment/comments/comment_form.html
```diff
@@ -8,8 +8,9 @@
...
 {% for field in comment_form.visible_fields %}
     {% if field.field.label == 'Content' %}
-    {% render_field field placeholder=placeholder class='js-comment-input form-control form-control-sm rounded-0' %}
+        {% render_field field placeholder=placeholder class='js-comment-input form-control form-control-sm rounded-0' %}
     {% else %}
-    {% render_field field class='js-comment-input form-control form-control-sm rounded-0' %}
+        {% render_field field class='js-comment-input form-control form-control-sm rounded-0' %}
     {% endif %}
 {% endfor %}
+<p class="error small text-danger mt-2 d-none">class</p>
```
abhiabhi94, line 16: is `'class'` written here inside the `p` tag a mistake or do i miss something?

comment/mixins.py
```diff
@@ -1,16 +1,40 @@
...
+            'error': self.error,
+            'anonymous': self.anonymous,
+            'msg': self.msg
+        }
+
+
+class UTF8JsonResponse(JsonResponse):
+    def __init__(self, *args, json_dumps_params=None, **kwargs):
+        json_dumps_params = {"ensure_ascii": False, **(json_dumps_params or {})}
+        super().__init__(*args, json_dumps_params=json_dumps_params, **kwargs)
```
abhiabhi94, line 37: maybe we can move these to a separate module `responses.py`.

comment/mixins.py
```diff
@@ -1,16 +1,40 @@
...
+    anonymous = False
+    msg = None
+
+    def json(self):
+        return {
+            'data': self.data,
+            'error': self.error,
+            'anonymous': self.anonymous,
+            'msg': self.msg
+        }
```
abhiabhi94, line 31: it could be more intuitive if we could rename this function to something along the lines of `response_json`, so that we get more clear information when we call it from a view.

comment/mixins.py
abhiabhi94, line 37: Maybe this looks more cleaner
```suggestion
    def __init__(self, *args, json_dumps_params={}, **kwargs):
        json_dumps_params["ensure_ascii"] =  False
        super().__init__(*args, json_dumps_params=json_dumps_params, **kwargs)

File: comment/validators.py (hunk 1 above)
Comment (Radi85) [reply comment/validators.py:137]: Hmmm, you have previously asked me to move the validitaor to a custom one 😝.
Using the custom validator is cleaner than writing try block inside the view and our vlidation message is identical to the default one, I added it to the message class so that it can be added to the custom translations.

File: comment/templates/comment/comments/comment_form.html (hunk 2 above)
Comment (Radi85) [reply comment/templates/comment/comments/comment_form.html:16]: 😄 it was for styling the error. I will remove it.

File: comment/mixins.py (hunk 3 above)
Comment (Radi85) [reply comment/mixins.py:37]: Seems eligible.

File: comment/mixins.py (hunk 4 above)
Comment (Radi85) [reply comment/mixins.py:31]: I am more convenient with `json` since the same is used in the default django response `response.json()`.

File: comment/mixins.py (hunk 3 above)
Comment (Radi85) [reply comment/mixins.py:37]: No it's not. this is more dangerous 😄 Never pass `{}` or `[]` as a default argument value.
[Read this](https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python)

File: comment/validators.py (hunk 1 above)
Comment (abhiabhi94) [reply comment/validators.py:137]: > I added it to the message class so that it can be added to the custom translations.

You can probably trust `django` to have more translation coverage than our app at the moment :smile:

File: comment/mixins.py (hunk 3 above)
Comment (abhiabhi94) [reply comment/mixins.py]: yes, yes, how could I forget, I have been bitten once. :)

**Comment by abhiabhi94** [comment]
I get an error when trying to post an anonymous comment from the UI, `Unable to post your comment!, please try again`.

Also, you probably forgot to change the response for `ConfirmComment` view.

File: comment/validators.py (hunk 1 above)
Comment (Radi85) [reply comment/validators.py:137]: 😒 I do. please read my statement again. I said that the message can be added to **custom translations.** So that the developer can choose the suitable translation and I believe django or any other app doesn't completely fit with the tastes of all people.

**Comment by Radi85** [comment]
> I get an error when trying to post an anonymous parent comment from the UI, `Unable to post your comment!, please try again`.

I am not sure that I can help with the current info since it's working for me locally.
I will try to test this branch again on a fresh project.

> Also, you probably forgot to change the response for `ConfirmComment` view.

 In `ConfirmComment` We redirect the user to the site after saving the comment so there is no need to use JsonResponse here.

**Comment by Radi85** [comment]
I will make new release `2.6b2` after merging this

**Comment by abhiabhi94** [comment]
> In ConfirmComment We redirect the user to the site after saving the comment so there is no need to use JsonResponse here.

Yes, you're right but there's also a case, where you render a template when the key passed doesn't resolve into a valid comment.

https://github.com/Radi85/Comment/blob/5f845cf85099eabcd0e4492f95f9c6db8cecb801/comment/views/comments.py#L125-L126

**Comment by Radi85** [comment]
> https://github.com/Radi85/Comment/blob/5f845cf85099eabcd0e4492f95f9c6db8cecb801/comment/views/comments.py#L125-L126

Oh nice observation, good you mentioned that. I will update it.

**Comment by abhiabhi94** [comment]
The GIF of the error when trying to post an anonymous comment locally on my machine.

anonymous-comment-error

I got the issue, it is occurring because of the `firefox relay icon` present in the email field.

I have used comments in the `javascript` code below to give more of an idea about the issue.
```js
...
/* here formButton actually captures the relay-icon */ 
 let formButton = form.querySelector("button");
...
/* Although it is a parent comment, since it doesn't get the attribute `value`, hence considers it as a child comment */
// parent comment
if (formButton.getAttribute('value') === 'parent') {
    // reload all comments only when posting parent comment
    if (result.anonymous) {
        createInfoElement(form.closest('.js-comment'), 'success', result.msg, 3);
        form.reset();
        return;
    }
    document.getElementById("comments").outerHTML = result.data;
} else {
// child comment
    if (result.anonymous) {
    /*Since it doesn't get the parent element here, it passed null to `createInfoElement` which was causing an error*/
    createInfoElement(form.closest('.js-parent-comment'), 'success', result.msg, 3);
    form.reset();
    return; 
    }
...
}
```


Although this might not occur with a lot of other people, maybe we can use a  more comprehensive check to use some sort of id or class rather than using the element name for the query to find the  button.

Radi85:
https://github.com/Radi85/Comment/blob/5f845cf85099eabcd0e4492f95f9c6db8cecb801/comment/views/comments.py#L125-L126

I don't think this needs to be changed to JsonResponse since this call is not handled in JS. The current response seems to be the right one.

Radi85:
> The GIF of the error when trying to post an anonymous comment locally on my machine.
> 
> anonymous-comment-error
> 
> I got the issue, it is occurring because of the `firefox relay icon` present in the email field.

I tested on Firefox and it's working! I didn't get what the issue is yet.

abhiabhi94:
> I tested on Firefox and it's working! I didn't get what the issue is yet.

Yes, it won't affect unless you use an extension [firefox-relay](https://addons.mozilla.org/en-US/firefox/addon/private-relay/). 

This generates email aliases that can be forwarded to your original email and ... 

So in my case it is causing an issue because it creates a button icon(to generate email aliases) as seen in the GIF above on the email fields.

Radi85:
> > I tested on Firefox and it's working! I didn't get what the issue is yet.
> 
> Yes, it won't affect unless you use an extension [firefox-relay](https://addons.mozilla.org/en-US/firefox/addon/private-relay/).
> 
> This generates email aliases that can be forwarded to your original email and ...
> 
> So in my case it is causing an issue because it creates a button icon(to generate email aliases) as seen in the GIF above on the email fields.

So this is a very special case. However, I am thinking to change the UI for creating anonymous comments to be similar to the one implemented in the subscription feature. That means the email field will be removed from the current UI form and instead will be prompted by a modal form.

abhiabhi94:
> However, I am thinking to change the UI for creating anonymous comments to be similar to the one implemented in the subscription feature. That means the email field will be removed from the current UI form and instead will be prompted by a modal form.

I think this should ideally be done in a separate issue.

PR #154, opened 2021-01-20
feat(#83): Add ordering option for comments

### LINKED ISSUE #83: top comments based on likes
Is there a way to list for example top 5 comments based on likes on a Post?

abhiabhi94:
closes #83


comment/conf/defaults.py
```diff
@@ -28,3 +28,5 @@
 COMMENT_ALLOW_TRANSLATION = False
 
 COMMENT_ALLOW_SUBSCRIPTION = False
+
+COMMENT_ORDER_BY = '-posted'
```
Radi85, line 32: Why are you limiting this to one value?
```suggestion
COMMENT_ORDER_BY = ['-posted']
```
We can use list of values and unpack it in `order_by`
https://docs.djangoproject.com/en/3.1/ref/models/querysets/#order-by

comment/managers/comments.py
```diff
@@ -16,8 +16,13 @@ def all_exclude_flagged(self):
 
         return super().get_queryset().exclude(flag__state__exact=2)
 
+    @staticmethod
+    def _filter_parents(qs, parent=None):
+        return qs.filter(parent=parent).order_by(validate_order())
```
Radi85, line 21: return list from `validate_order()`
```suggestion
        return qs.filter(parent=parent).order_by(*validate_order())
```

comment/utils.py
```diff
@@ -206,3 +207,24 @@ def get_username_for_comment(comment):
             return comment.email.split('@')[0]
         return settings.COMMENT_ANONYMOUS_USERNAME
     return comment.user.username
+
+
+def validate_order():
+
+    @lru_cache()
+    def get_allowed_orders():
```
Radi85, line 215: Can you please move this method outside. And please avoid using nested functions except for docerators.

Radi85 (changes requested):
I will have a deeper look tomorrow.

comment/utils.py
abhiabhi94: the method `get_allowed_orders` isn't required anywhere else in the codebase. To avoid exposing it anywhere else, it is used here. Such types of closures are even used in the official python codebase.

Using classes here seems an overkill here to me.

comment/conf/defaults.py
abhiabhi94: valildation could get a bit tricky, and it would be tough to explain the documentation, that is why i avoided this at first.

Suppose, we allow a tuple and check if each value of the tuple is present in allowed ones, something like this would also escape this: `('posted', '-posted')` and many more like these. 

Linking to the above `django` documentation(one that you pointed) is an option, but they allow a lot of other things such as model's relational fields, query expressions, etc.

Another option is just to ignore this. This sort of thing isn't quiet explained well even in the `django` documentation(at least I couldn't see it). 

I couldn't quite think of an elegant way to solve this. If you have suggestions it would be great to know them.

comment/utils.py
```diff
@@ -206,3 +207,24 @@ def get_username_for_comment(comment):
...
+        return ALLOWED_ORDERS
+
+    allowed_orders = get_allowed_orders()
+    prefered_order = settings.COMMENT_ORDER_BY
+    if prefered_order in allowed_orders:
+        return prefered_order
+
+    raise ImproperlyConfigured(
+            ErrorMessage.INVALID_ORDER_ARGUMENT.format(order=prefered_order, allowed_orders=allowed_orders)
+        )
```
Radi85, line 230:
```suggestion
@lru_cache()
def get_allowed_orders():
    allowed_orders = ['reaction__likes', 'reaction__dislikes', 'posted']
    # adds support for descending order as well
    allowed_orders.extend(list(map(lambda a: '-' + a, allowed_orders)))
    # django allows this to support random order
    allowed_orders.append('?')
    return allowed_orders


def validate_order():
    preferred_orders = settings.COMMENT_ORDER_BY
    unique_values = set(map(lambda a: a.replace('-', ''), preferred_orders))
    if len(unique_values) != len(preferred_orders):
        duplicated_orders = list(set(preferred_orders) - unique_values)
        raise ImproperlyConfigured(
            ErrorMessage.DUPLICATE_ORDER_VALUE.format(
                duplicates=duplicated_orders, order=duplicated_orders[0].replace('-', '')
            )
        )

    allowed_orders = get_allowed_orders()
    for preferred_order in preferred_orders:
        if preferred_order not in allowed_orders:
            raise ImproperlyConfigured(
                ErrorMessage.INVALID_ORDER_ARGUMENT.format(order=preferred_order, allowed_orders=allowed_orders)
            )

    return preferred_orders
```

This is what came up with me. I tested and it's working fine with order list. Feel free to refactor it and please pay attention to this:
- Please use lowercase for variable name `ALLOWED_ORDER`.
- Please don't use nested function. I will show you my reason in the following comment.
- I used this validation message:
```python
    DUPLICATE_ORDER_VALUE = _((
        'Comment app: COMMENT_ORDER_BY should not have duplicated values '
        'Duplicated Values: {duplicates}. Please use one value only E.g. "{order}" or "-{order}".'
    ))
```

Radi85 (changes requested):

comment/utils.py
Radi85: My reason to not use nested functions is to avoid classing/mixing scope variables. This screenshot elabroates.
Screenshot 2021-01-23 at 09 31 41

You can clearly see how the variables clash.

abhiabhi94: > Please use lowercase for variable name ALLOWED_ORDER.

It is a constant, and it is advisable to use all capital letters for such cases. You may see the PEP8 guidelines https://www.python.org/dev/peps/pep-0008/#id48

>  DUPLICATE_ORDER_VALUE = _((
        'Comment app: COMMENT_ORDER_BY should not have duplicated values '
        'Duplicated Values: {duplicates}. Please use one value only E.g. "{order}" or "-{order}".'
    ))

Disallowing duplicate values seems like moral policing to me. This thing is syntactically valid and throws up some result. Just because we can't explain the result, should we disallow it?

abhiabhi94: > My reason to not use nested functions is to avoid classing/mixing scope variables. This screenshot elabroates.

You have used a particular example. I agree that it wouldn't have been a great practice in that particular case. 

In the current scenario, I don't think any variables from outer scopes were used inside the nested function. Also, in this case, it also prevents exposing the nested function to any other part of the code as they don't necessarily need to do anything with it.

Trying to extrapolate and generalize the issue caused in the screenshot to every example is an overreaction and maybe a classic of [faulty generalization](https://en.wikipedia.org/wiki/Faulty_generalization).

If it were that true, surely they might have not used this in python codebase or mentioned some sort of warning to **never** use nested functions in maybe at lease one of the PEPs.

Radi85: > It is a constant, and it is advisable to use all capital letters for such cases. You may see the PEP8 guidelines https://www.python.org/dev/peps/pep-0008/#id48

This is valid for module variables/constants but it is not for functions and methods scope variables.

PEP8 statement:
> Constants are usually defined on a module level and written in all capital letters with underscores separating words. Examples include MAX_OVERFLOW and TOTAL.

---------------
> Disallowing duplicate values seems like moral policing to me. This thing is syntactically valid and throws up some result. Just because we can't explain the result, should we disallow it?

I didn't get what you refer to!

abhiabhi94: > I didn't get what you refer to!

As mentioned in your suggested code, you're trying to disallow duplicated values passed to the list and hence the error message regarding duplication.

> This is valid for module variables/constants but it is not for functions and methods scope variables.

Okay, so moving this to a module level constant should be a better thing. It might also settle the debate on the use of nested functions as of now.

Radi85: > You have used a particular example. I agree that it wouldn't have been a great practice in that particular case.
> 
> In the current scenario, I don't think any variables from outer scopes were used inside the nested function. Also, in this case, it also prevents exposing the nested function to any other part of the code as they don't necessarily need to do anything with it.

My statement is general, it's not related to a particular example. using nested function will most likely cause a variables clashing. Why would I need to prevent exposing such a method? What's wrong with that? 🤔 

> Trying to extrapolate and generalize the issue caused in the screenshot to every example is an overreaction and maybe a classic of [faulty generalization](https://en.wikipedia.org/wiki/Faulty_generalization).

Take it easy. Your language seems to be aggressive here.

> If it were that true, surely they might have not used this in python codebase or mentioned some sort of warning to **never** use nested functions in maybe at lease one of the PEPs.

Also you are mixing things up, nested functions are needed for some and special use cases, why would PEP inform us to not use them?!!

Radi85: > As mentioned in your suggested code, you're trying to disallow duplicated values passed to the list and hence the error message regarding duplication.

Yes duplication should not be allowed, right?

> Okay, so moving this to a module level constant should be a better thing. It might also settle the debate on the use of nested functions as of now.

I didn't say, move the variable to the module level!! I said use lowercase name for variables in the function's scope.
Uppercase names are used for module variables as per PEP8 and THIS IS NOT OUR CASE.

abhiabhi94: > Take it easy. Your language seems to be aggressive here.

I'm sorry if it came out that way, I was just trying out to point out something that seemed as a potential [logical fallacy](https://en.wikipedia.org/wiki/Formal_fallacy) to me in your argument.

> My statement is general, it's not related to a particular example. using nested function will most likely cause a variables clashing.

In the example that you show in the screenshot the variable `request` is reused from an external scope and I agree it is not a good practice.
But in the function below, I don't see it as using any external variable, so how is it clashing with anything else. 
```python
lru_cache()
def get_allowed_orders():
    allowed_orders = ['reaction__likes', 'reaction__dislikes', 'posted']
    # adds support for descending order as well
    allowed_orders.extend(list(map(lambda a: '-' + a, allowed_orders)))
    # django allows this to support random order
    allowed_orders.append('?')
    return allowed_orders
```

>  Why would I need to prevent exposing such a method? What's wrong with that? thinking

This is what encapsulation is all about? This method is an internal part of implementation of logic of validating `order_by` parameter. Please correct if I'm wrong.

> Yes duplication should not be allowed, right?

Duplication is a valid in the query. 

```sh
$ python manage.py shell         
>>> from comment.models import Comment                                                     
>>>
>>> Comment.objects.all().order_by('posted', '-posted')                                    
<QuerySet [comment by moderator: First comment, reply by test: Reply comment, comment by admin: Second comment, comment by a: hi, comment by tester: new comment, comment by tester: hey, reply by tester: hi, reply by admin: hi, comment by admin: bla bla bals kdsjfds, reply by admin: hi]>
>>>
>>> Comment.objects.all().order_by('posted', '-posted', 'posted')                          
<QuerySet [comment by moderator: First comment, reply by test: Reply comment, comment by admin: Second comment, comment by a: hi, comment by tester: new comment, comment by tester: hey, reply by tester: hi, reply by admin: hi, comment by admin: bla bla bals kdsjfds, reply by admin: hi]>
```

> I didn't say, move the variable to the module level!! I said use lowercase name for variables in the function's scope.
Uppercase names are used for module variables as per PEP8 and THIS IS NOT OUR CASE.

I think we can move the variable to a module level constant because it is essentially a constant, and mark it as private since it may change with time.

Radi85: > I'm sorry if it came out that way, I was just trying out to point out something that seemed as a potential [logical fallacy](https://en.wikipedia.org/wiki/Formal_fallacy) to me in your argument.

No worries. Just wanted to make sure nothing disturbed you 😄 


> In the example that you show in the screenshot the variable `request` is reused from an external scope and I agree it is not a good practice.
> But in the function below, I don't see it as using any external variable, so how is it clashing with anything else.
> 
> ```python
> lru_cache()
> def get_allowed_orders():
>     allowed_orders = ['reaction__likes', 'reaction__dislikes', 'posted']
>     # adds support for descending order as well
>     allowed_orders.extend(list(map(lambda a: '-' + a, allowed_orders)))
>     # django allows this to support random order
>     allowed_orders.append('?')
>     return allowed_orders
> ```

Again, I'm generalising this issue. What if someone needed a variable name `allowed_orders` in the outer function without realizing that it's already used in the nested function? Debugging and time consuming.

I personally faced such a problem and it took me hours debugging to find that there are variable names clashing in nested scopes. My idea is to avoid these expected issues by moving the function to its own scope.

> This is what encapsulation is all about? This method is an internal part of implementation of logic of validating `order_by` parameter. Please correct if I'm wrong.

The above case voilates the encapsulation more. 😃 

> Duplication is a valid in the query.

Right, In fact you raised the duplication issue in a previous comment and I also think it's better to restrict to not repeating the value for 2 reasons:

- Duplicating the value is useless and might give unnoticeable unexpected behaviour. I am not sure which value/order will be considered here `.order_by('posted', '-posted')`.
- We are using a settings variable so it would be nice to be checked for a reasonable value and give the developer feedback on their value while development.

> I think we can move the variable to a module level constant because it is essentially a constant, and mark it as private since it may change with time.

You are using a function to manipulate this variable so I don't think the you need to use module constant variable.

abhiabhi94: > Right, In fact you raised the duplication issue in a previous comment and I also think it's better to restrict to not repeating the value for 2 reasons:
    Duplicating the value is useless and might give unnoticeable unexpected behaviour. I am not sure which value/order will be considered here .order_by('posted', '-posted').
    We are using a settings variable so it would be nice to be checked for a reasonable value and give the developer feedback on their value while development.

Who are we to restrict them to something when `django` allows it? I agree the results can be weird but if it were not stopped by `django` maybe we should also consider allowing them.

If we want to be too opinionated(in my opinion), we can give warnings to the developer for duplicate values.

Radi85: > If we want to be too opinionated(in my opinion), we can give warnings to the developer for duplicate values.

This is all about. How to give the developer warning?!

Keep in mind that if X or Y uses a particular method or a specific lifestyle, that doesn't mean it fits your style or your need but it just fits theirs.

abhiabhi94: > How to give the developer warning?!

Using [`RunTimeWarning`](https://docs.python.org/3/library/exceptions.html#RuntimeWarning)

> Keep in mind that if X or Y uses a particular method or a specific lifestyle, that doesn't mean it fits your style or your need but it just fits theirs.

I'm afraid I again feel this to be a generalization of my statement. Let me try to clarify a bit. 
We are building an app/library that adds functionality to `django`. So in my opinion, it don't think it would be great to restrict something that `django` allows.

Radi85: > I'm afraid I again feel this to be a generalization of my statement. Let me try to clarify a bit.
> We are building an app/library that adds functionality to `django`. So in my opinion, it don't think it would be great to restrict something that `django` allows.

Do you know the reason behind allowing this? Are you sure it's not a bug? In the end Django is not a Holy Book.

abhiabhi94: > Do you know the reason behind allowing this? Are you sure it's not a bug?

Good point, as you can in see the raw SQL queries below, duplication isn't actually passed to the database.

```python
>>> print(Comment.objects.all().order_by('posted', '-posted').query)                        
SELECT "comment_comment"."id", "comment_comment"."user_id", "comment_comment"."email", "comment_comment"."parent_id", "comment_comment"."content_type_id", "comment_comment"."object_id", "comment_comment"."content", "comment_comment"."urlhash", "comment_comment"."posted", "comment_comment"."edited" FROM "comment_comment" ORDER BY "comment_comment"."posted" ASC

>>> print(Comment.objects.all().order_by('-posted', 'posted').query)                        
SELECT "comment_comment"."id", "comment_comment"."user_id", "comment_comment"."email", "comment_comment"."parent_id", "comment_comment"."content_type_id", "comment_comment"."object_id", "comment_comment"."content", "comment_comment"."urlhash", "comment_comment"."posted", "comment_comment"."edited" FROM "comment_comment" ORDER BY "comment_comment"."posted" DESC

# even for more than two values they consider just the first value of the fields.
>>> print(Comment.objects.all().order_by('-posted', 'posted', 'posted').query)              
SELECT "comment_comment"."id", "comment_comment"."user_id", "comment_comment"."email", "comment_comment"."parent_id", "comment_comment"."content_type_id", "comment_comment"."object_id", "comment_comment"."content", "comment_comment"."urlhash", "comment_comment"."posted", "comment_comment"."edited" FROM "comment_comment" ORDER BY "comment_comment"."posted" DESC

>>> print(Comment.objects.all().order_by('-posted', 'posted', '-posted').query)             
SELECT "comment_comment"."id", "comment_comment"."user_id", "comment_comment"."email", "comment_comment"."parent_id", "comment_comment"."content_type_id", "comment_comment"."object_id", "comment_comment"."content", "comment_comment"."urlhash", "comment_comment"."posted", "comment_comment"."edited" FROM "comment_comment" ORDER BY "comment_comment"."posted" DESC
```

I will try to see more in their code and see what more I can gather.

comment/validators.py
```diff
@@ -135,3 +140,37 @@ def is_valid(self):
...
+    if len(unique_values) != len(preferred_orders):
+        duplicated_orders = list(set(preferred_orders) - unique_values)
+        raise ImproperlyConfigured(
+            ErrorMessage.DUPLICATE_ORDER_VALUE.format(
+                duplicates=duplicated_orders, order=duplicated_orders[0].replace('-', '')
+            )
+        )
+
+
+def _validate_order():
```
Radi85, line 166: Any reason for making/tagging this function as private `_validate_order` despite it's intended to be used outside the module?

Radi85 (changes requested):

comment/validators.py
```diff
@@ -135,3 +140,37 @@ def is_valid(self):
...
+def _get_allowed_orders():
+    allowed_orders = ['reaction__likes', 'reaction__dislikes', 'posted']
+    # adds support for descending order as well
+    allowed_orders.extend(list(map(lambda a: '-' + a, allowed_orders)))
+    # django allows this to support random order
+    allowed_orders.append('?')
+    return allowed_orders
+
+
+def _validate_orders_are_unique(preferred_orders):
```
Radi85, line 155: Wouldn't be this name more eligible:
```suggestion
def _validate_unique_orders(preferred_orders):
```

Radi85 (changes requested):

comment/validators.py
abhiabhi94: internal functions too can be prefixed with  `_`. Just to remind people that this method may not be backward compatible in the future. I think we have gone over this stuff in the past  as well.

comment/validators.py
abhiabhi94: validate orders are unique seems more grammatically correct to me over validate unique orders. To me, the former seems to not do something that we intended.

Radi85: Are we writing a novel? 😄 In fact I usually don't see verb `to be` inside the function names.

abhiabhi94: on reading validate unique orders, the first though that comes to my mind is this function must be doing is validating orders which are unique(i.e. already unique orders have been passed to it, and it is performing some validation over them), while what we intended to do is to validate whether orders are unique or not.

Radi85: I do agree with you but this name `_validate_unique_orders` what I usally see for this kind of function/methods. Feel free to keep your naming.

abhiabhi94: I think `are_orders_unique` seems to be striking the right chords to me. Right?

abhiabhi94: on second thoughts, doesn't seem to be a great choice since `are_orders_unique` should just be returning `bool` and not raising exceptions, if not.

comment/validators.py
Radi85: I am not sure this fits to our case. Our app is just a plug-in not a framework.

abhiabhi94: Since we are exposing our code to the client, this is just a way to tell them not to rely on this function. It is for internal use and any changes can be made in into it without a major version bump.

Radi85: Where did you get this info from?
Can you please share this info?
I personally didn't see frameworks prefix their methods and functions with `_` to let people know that it is not backward compatible, otherwise we will see frameworks full of prefixed `_` functions.
Also this is developers invention not a language built-in and it's evantually up to the user's preference and as my preference I don't like to use them. :)

abhiabhi94: > Where did you get this info from?
Can you please share this info?

I'm also quoting the relevant sections here.

Python: https://docs.python.org/3/tutorial/classes.html#tut-private
>> “Private” instance variables that cannot be accessed except from inside an object don’t exist in Python. However, there is a convention that is followed by most Python code: a name prefixed with an underscore (e.g. _spam) should be treated as a non-public part of the API (whether it is a function, a method or a data member). It should be considered an implementation detail and subject to change without notice.

Django: https://docs.djangoproject.com/en/dev/misc/api-stability/#apis-marked-as-internal

>> Certain APIs are explicitly marked as “internal” in a couple of ways:
    Some documentation refers to internals and mentions them as such. If the documentation says that something is internal, we reserve the right to change it.
    Functions, methods, and other objects prefixed by a leading underscore (_). This is the standard Python way of indicating that something is private; if any method starts with a single _, it’s an internal API.

If I remember correctly, we have gone over this earlier, when there was a discussion around using ` _get_version` with a leading `_` inside the `__init__` module. 


> Also this is developers invention not a language built-in and it's evantually up to the user's preference and as my preference I don't like to use them. :)

Yes, this is developer invention and so, is the complete language. I like this convention because it explicitly tells the users about the stability of the API.

Radi85: Cool, Thanks for sharing. So let's use this convention.

Radi85:
Hey @abhiabhi94,
Can you please rebase this PR again?

Radi85 (approved):

PR #155, opened 2021-01-31
feat(#117): Add support for customization of user fields in the api

### LINKED ISSUE #117: Better UserSerializer

abhiabhi94:
- resolves #117
- closes #126

### DIFF
--- a/comment/api/serializers.py
+++ b/comment/api/serializers.py
@@ -2,6 +2,7 @@ from django.contrib.auth import get_user_model
 from django.utils import timezone
 
 from django.apps import apps
+from django.db.models import ImageField
 
 from rest_framework import serializers
 
@@ -23,10 +24,12 @@ def get_profile_model():
 def get_user_fields():
     user_model = get_user_model()
     fields = user_model._meta.get_fields()
+    api_fields = list(settings.COMMENT_USER_API_FIELDS) + ['profile']
+    api_fields = list(set(api_fields))
     for field in fields:
-        if hasattr(field, "upload_to"):
-            return 'id', 'username', 'email', 'profile', field.name
-    return 'id', 'username', 'email', 'profile'
+        if hasattr(field, "upload_to") and isinstance(field, ImageField):
+            api_fields.append(field.name)
+    return api_fields
 
 
 class ProfileSerializerDAB(serializers.ModelSerializer):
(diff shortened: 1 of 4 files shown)


comment/api/serializers.py
```diff
@@ -23,10 +26,13 @@ def get_profile_model():
 def get_user_fields():
     user_model = get_user_model()
     fields = user_model._meta.get_fields()
+    api_fields = list(settings.COMMENT_USER_API_FIELDS) + ['profile']
+    # remove duplicates, preserve order
+    api_fields = list(OrderedDict.fromkeys(api_fields))
```
Radi85, line 31: Please use `set` for removing duplication.
- More readability.
- No extra import statement needed.
- No comment needed to explain what this line is doing.

```suggestion
    api_fields = list(set(api_fields))
```

Radi85 (changes requested):

comment/api/serializers.py
abhiabhi94: `sets` don't necessarily prevent orders and that is what the comment says. we would want to preserve the order in which custom fields have been specified.

Radi85: I am not sure that I understood what you mean, but why would we care about the orders? We are getting unique values from a list, right?

abhiabhi94: i have set the fields to `[`a`, 'b']` and now my frontend code is written in a way that tends to care about the order in which these fields are returned.

Radi85: Hmmm! You might be confused here? The returned value is json not an ordered list, so these fields will be used in json object like so:
```python
fields = ['a', 'b']

response = {
    'a': a_value,
    'b': b_value
}
```

That means keeping the same order has no meaning.

Radi85 (changes requested):
Docs need to be updated accordingly!

PR #164, opened 2021-02-11
feat(#128): Allow custom user model fields

### LINKED ISSUE #128: Use django built-in EMAIL_FIELD and USERNAME_FIELD
### Change Type
[x] Refactor/Enhancement
[ ] Testing
[ ] CI/CD
[ ] Other

### Proposed Changes
Replace all occurrences of `user.username` and `user.email` with django built-in fields `USERNAME_FIELD` and `EMAIL_FIELD` respectively.

This will allow us to use the actual `username` and `email` fields on the user model instead of using the default one.

abhiabhi94:
- custom username and email fields set inside user model can now be used.

resolves #128

### DIFF
--- a/comment/models/comments.py
+++ b/comment/models/comments.py
@@ -8,7 +8,7 @@ from django.utils import timezone
 
 from comment.managers import CommentManager
 from comment.conf import settings
-from comment.utils import is_comment_moderator, get_username_for_comment
+from comment.utils import is_comment_moderator
 
 
 class Comment(models.Model):
@@ -33,11 +33,12 @@ class Comment(models.Model):
         ordering = ['-posted', ]
 
     def __str__(self):
-        username = get_username_for_comment(self)
+        username = self.get_username()
+        _content = self.content[:20]
         if not self.parent:
-            return f'comment by {username}: {self.content[:20]}'
+            return f'comment by {username}: {_content}'
         else:
-            return f'reply by {username}: {self.content[:20]}'
+            return f'reply by {username}: {_content}'
 
     def __repr__(self):
         return self.__str__()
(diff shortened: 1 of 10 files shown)


comment/models/comments.py
```diff
@@ -8,12 +8,12 @@
...
 from comment.managers import CommentManager
 from comment.conf import settings
-from comment.utils import is_comment_moderator, get_username_for_comment
+from comment.utils import is_comment_moderator
 
 
 class Comment(models.Model):
     user = models.ForeignKey(get_user_model(), on_delete=models.CASCADE, blank=True, null=True)
-    email = models.EmailField(blank=True)
+    email = models.EmailField(blank=True, null=True)
```
Radi85, line 16: What the purpose of this change?

Radi85 (changes requested):

comment/models/comments.py
abhiabhi94: to allow email field to accept null values. `django` allows email fields to have null values. this point was discussed earlier in https://github.com/Radi85/Comment/issues/128#issuecomment-751258182 and also, in one of the pull request that didn't get merged.

Radi85: I mean we didn't have any issue with the current state, why would we need to do this change now? I think you can replace None with empty string to avoid new migration.
The chabne should be in the _set_email method:
https://github.com/Radi85/Comment/pull/164/files#diff-c4e5567003cbea1eddb9e40211f2c07e1608ac78a0a54f221d18b0332ed36a2fL75-L78
```python
self.email = getattr(self.user, self.user.EMAIL_FIELD, '')
```

abhiabhi94: > I mean we didn't have any issue with the current state, why would we need to do this change now?

To avoid future instances of such cases where people aren't using email but only usernames. There are a lot more scenarios for this.

> I think you can replace None with empty string to avoid new migration.

I don't understand what's so concerning about the migration as to say. Also, unless and until we have `null=True` on the model field, how can we set an empty string for it? An `EmailField` won't allow those value.

Radi85: > To avoid future instances of such cases where people aren't using email but only usernames. There are a lot more scenarios for this.

You are checking if the user object has an email attribute and you use it if any, otherwise you are using the default value from `getattr`, what scenarios can be other than this???

> I don't understand what's so concerning about the migration as to say. 

Avoid adding unnecessary code specially migrations.

> Also, unless and until we have `null=True` on the model field, how can we set an empty string for it? An `EmailField` won't allow those value.

We didn't have any issue so far on creating comment with an empty email value. The user can still create a comment even if their email value is empty.

Radi85 (approved):

PR #169, opened 2021-02-27
feat(#158): allow rendering new line in comment

### LINKED ISSUE #158: Allow users to make new paragraphs in comments
### Proposed Functionality
In the comments currently, the library doesn't allow the users to make a new comment by hitting enter twice- everything instead is on a single line.  Can you please make it so that when a user makes a comment in the preview, and tries to make a new paragraph, it also shows up when they post it as well? 

Preview of comment
Screen Shot 2021-02-03 at 9 20 00 AM

What the comment look like when the user makes a post
Screen Shot 2021-02-03 at 9 20 13 AM

Radi85:
closes #158


comment/templatetags/comment_tags.py
```diff
@@ -82,9 +82,13 @@ def render_comments(obj, request, oauth=False):
...
-def render_content(comment, number):
-    number = int(number)
-    content = comment.content
+def render_content(comment, number=None):
+    try:
+        number = int(number)
+    except (ValueError, TypeError):
+        number = get_wrapped_words_number()
+
+    content = comment.content.replace('\n', ' <br/> ')
```
abhiabhi94, line 91: I think this won't be the most appropriate solution for long strings where the user puts more than 1 newline-character. Something like:

```html
Something ends here.(<b>notice the use of 2 new-line characters</b>.)


Something new starts here.
```
will get transformed to:
```html
Something ends here.<br><br>Something starts here.
```
 while we might want this to be replaced as: `Something ends here.<br>Something new starts here.`

You may use:
```python
import re
# this can be put at the top of the module
NEW_LINE_RE = re.compile(r'(.*)(\n)+(.*)')
content = NEW_LINE_RE.sub(r'\1<br>\3', comemnt.content)
```

abhiabhi94:
Hey @Radi85, I have a couple of suggestions:

- Maybe we can update the `changelog` alongside the changes, so that it is ready when a new version is released.
- Instead of the working on a good-first kind of issue(unless it is a critical one), maybe we can remove the milestone and wait to see if someone from the community is willing enough to work on it. This will help us also in getting some sort of a feedback, whether the feature/issue is really required by users.

comment/templatetags/comment_tags.py
Radi85: I believe we should leave the number of lines to the user, they might intend to insert multiple breaks, why would we restrict it to one break?

Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book. It has survived not only five centuries.



but also the leap into electronic typesetting, remaining essentially unchanged. It was popularised in the 1960s with the release of Letraset sheets containing Lorem Ipsum passages, and more recently with desktop publishing software like Aldus PageMaker including versions of Lorem Ipsum

abhiabhi94: > I am just thinking, that we should restrict the number otherwise the comment style can be screwed.

when wrapping content, it seems more reasonable to me replace all line-breaks by one and then replace it with a space character. this is generally what i see in most blogging types of platforms when they do this sort of thing for "list views".

Also, i am not sure this `\n` approach is going to work for languages other than English. I actually peeked into the django code to look for how they use the `[truncatechars](https://docs.djangoproject.com/en/3.1/ref/templates/builtins/#truncatechars)` filter. They happen to use a lot of processing around `unicodes` under the hood.

I'm just pasting the relevant portion of the code: https://github.com/django/django/blob/a948d9df394aafded78d72b1daa785a0abfeab48/django/utils/text.py#L102-L117

 ```python
 def _text_chars(self, length, truncate, text, truncate_len):
        """Truncate a string after a certain number of chars."""
        s_len = 0
        end_index = None
        for i, char in enumerate(text):
            if unicodedata.combining(char):
                # Don't consider combining characters
                # as adding to the string length
                continue
            s_len += 1
            if end_index is None and s_len > truncate_len:
                end_index = i
            if s_len > length:
                # Return the truncated string
                return self.add_truncation_text(text[:end_index or 0],
                                                truncate)
```


If you want you can look through the actual code: https://github.com/django/django/blob/master/django/template/defaultfilters.py#L265

Comment (Radi85) [reply comment/templatetags/comment_tags.py:91]: I am not fully understanding what you are trying to achieve here. In our case we are truncating words not chars and django is using split method to get the word length as we did in our code:
https://github.com/django/django/blob/a948d9df394aafded78d72b1daa785a0abfeab48/django/utils/text.py#L134
```python
    def _text_words(self, length, truncate):
        """
        Truncate a string after a certain number of words.
        Strip newlines in the string.
        """
        words = self._wrapped.split()
        if len(words) > length:
            words = words[:length]
            return self.add_truncation_text(' '.join(words), truncate)
        return ' '.join(words)
```
Regarding `\n` linebreak I suppose this applies to all langauges not english only.

Screenshot 2021-03-05 at 10 42 30

Radi85:
> Hey @Radi85, I have a couple of suggestions:
> 
> * Maybe we can update the `changelog` alongside the changes, so that it is ready when a new version is released.

Changelog is updated on each pre/release. I don't mind to be updated on each change when necessary.

> * Instead of the working on a good-first kind of issue(unless it is a critical one), maybe we can remove the milestone and wait to see if someone from the community is willing enough to work on it. This will help us also in getting some sort of a feedback, whether the feature/issue is really required by users.

I agree.

comment/templatetags/comment_tags.py
abhiabhi94: > Regarding \n linebreak I suppose this applies to all langauges not english only.

Don't have an issue if it works.

> I am not fully understanding what you are trying to achieve here. In our case we are truncating words not chars and django is using split method to get the word length as we did in our code:

There is also a similar function for wrapping words and chars in django. My concern was not all languages are written from left to write, so this type of slicing words from the right might not be an ideal thing in that scenario. 
But, we can continue the current process and change it if an issue arises.

Regarding wrapping of content, in the current style, I think it would be necessary.

Radi85: I already tested the words wrapping for Arabic which is RTL and it seems to be working as expected. I believe the UI style for RTL should be our concern.

comment/templatetags/comment_tags.py
```diff
@@ -82,9 +84,15 @@ def render_comments(obj, request, oauth=False):
...
-    content = comment.content
+def render_content(comment, number=None):
+    try:
+        number = int(number)
+    except (ValueError, TypeError):
+        number = get_wrapped_words_number()
+
+    # Restrict 2 or more line breaks to 2 <br>
+    content = re.sub(r'(.*)(\n){2,}(.*)', r'\1<br><br>\3', comment.content)
+    content = re.sub(r'(.*)(\n)(.*)', r'\1<br>\3', content)
```
abhiabhi94, line 95: This will increase the time considerably because the same regular expression will be compiled again and again for all comments. Instead, you can compile the regular expression at the top of the module and use it here.
```suggestion
# use this at the top of the module
SINGLE_NEW_LINE_RE =  re.compile(r'(.*)(\n){2,}(.*)')
MORE_THAN_ONE_NEW_LINE_RE = re.compile(r'(.*)(\n)(.*)')

content = SINGLE_NEW_LINE_RE.sub(r'\1<br><br>\3', comment.content)
content = MORE_THAN_ONE_NEW_LINE_RE.sub(r'\1<br>\3', content)

```

abhiabhi94 (changes requested):

comment/templatetags/comment_tags.py
abhiabhi94: As far as I can see in the images that you have posted above, i can't see the wrapping in Arabic(i can see read less in Hindi but no such thing in Arabic)(I have assumed it is the first one). 

I don't have much knowledge about how right to left languages are actually read, so please pardon my naivety here.

Moreover, I'm just trying to demonstrate what my concern was through an example: 
```
>>> a = 'hi! this text is to be read from left to right'
>>> n = ''.join(list(reversed(a)))
>>> n
'thgir ot tfel morf daer eb ot si txet siht !ih'

# first 5 words
>>> ' '.join(n.split()[:5])
'thgir ot tfel morf daer'
```

Is this the expected result?

Radi85: No this is not same in RTL
```python
>>> s = 'تسجّل 1 الآن 2 لحضور المؤتمر الدولي العاشر ليونيكود، الذي سيعقد بمدينة مَايِنْتْس، ألمانيا 12'
>>> s.split()[:5]
['تسجّل', '1', 'الآن', '2', 'لحضور']
```

abhiabhi94: okay, so python does this by itself?, as far as I can see these are the last 5 words.

Radi85: Don't worry, I believe python is smart enough to know from where start counting the words 😉

abhiabhi94 (approved):

PR #170, opened 2021-02-27
fix(#157): pluralization issue

### LINKED ISSUE #157: Russian translate
### Environment
* Python version: 3.8.2

### Proposed Functionality
I am translating your module into Russian, and I cannot correctly organize the substitution of plural endings to words, since there are 3 forms of endings for plurals in Russian, it is not difficult to write this into a formula, but I do not understand where to write the endings themselves, except for changing the library's source code, to be exact the line below:

https://github.com/Radi85/Comment/blob/d2247a68dd624cc36a87f89cc2ea627bb6b44590/comment/templates/comment/comments/comment_content.html#L39

### Database Changes
No

Radi85:
closes #157


abhiabhi94 (approved):
It would be nice to not have this unnecessary overhead of translation files.

PR #178, opened 2021-03-25
fix: Toggling of subscription button in js

### LINKED ISSUE #175: Toggling of follow button in javascript breaks replies when subscription is disabled
### Environment
* python version = 3.9
* django Version = 3.1.1
* django-comments-dab = 2.6.0

### Describe the bug
I got a JS alert error "Unable to post your comment! please try again" when replying to a comment, after clicking the okay button the reply appended, while the form not clear

### Steps To Reproduce
1. Do not use the subscription option i.e. make sure `COMMENT_ALLOW_SUBSCRIPTION` is not set to `True`(which is the default behavior).
2. Try to reply to a comment.

### Expected behavior
The reply should submit without an error. 

### Screenshots
Capture

abhiabhi94:
- toggling only takes place when the button is present

fixes #175

### DIFF
--- a/comment/static/js/comment.js
+++ b/comment/static/js/comment.js
@@ -165,8 +165,10 @@ document.addEventListener('DOMContentLoaded', () => {
                 commentCount(1);
                 // update followBtn
                 let followButton = form.parentElement.previousElementSibling.querySelector(".js-comment-follow");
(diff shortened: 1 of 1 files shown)


abhiabhi94:
This can't be merged before #177 fixes CI.

Issue #147, opened 2021-01-10
Update setup documentation. 


Radi85:
### Change Type
[x] Addition
[ ] Correction
[ ] Deprecation
[ ] Cleanup (formatting, typos, etc.)

### Section
[ ] Installation
[x] Setup
[ ] Usage
[ ] REST API
[ ] Styling
[ ] Settings
[ ] Contribution/Development
[ ] Other

### Proposed Changes

Update instruction for adding `get_absolute_url` method to the model class that uses the comment model.

abhiabhi94:
I think we can add some labels like `good-first-issue` on easier issues like these to maybe encourage more contributions from the community.

PR #181, opened 2021-03-26
doc(#147): update setup documentation

ZendaInnocent:
- Add `get_absolute_url` method to the model class

### DIFF
--- a/docs/source/setup.rst
+++ b/docs/source/setup.rst
@@ -19,9 +19,21 @@ E.g. ``Post`` model, as shown below:
     class Post(models.Model):
         author = models.ForeignKey(User)
         title = models.CharField(max_length=200)
+        slug = models.SlugField(unique=True)
         body = models.TextField()
         # the field name should be comments
         comments = GenericRelation(Comment)
+
+.. important::
+
+    Please add a ``get_absolute_url`` method to the model that you're associating with the ``Comment`` model. It is used at several places.
+
+    .. code:: python
+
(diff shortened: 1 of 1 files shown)


docs/source/setup.rst
```diff
@@ -13,16 +13,21 @@ E.g. ``Post`` model, as shown below:
...
         author = models.ForeignKey(User)
         title = models.CharField(max_length=200)
+        slug = models.SlugField(unique=True)
         body = models.TextField()
         # the field name should be comments
         comments = GenericRelation(Comment)
 
+        def get_absolute_url(self):
+            return reverse('post_detail_url', kwargs={'slug': self.slug})
+
```
abhiabhi94, line 30: Maybe you can add a note that emphasizes that adding this method is important. Something of the form below seems more useful to me.

```rst
.. important:: 

    Please add a ``get_absolute_url`` method to the model that you're associating with the ``Comment`` model. It is used at several places.

    .. code:: python

        from django.urls import reverse
        
        def get_absolute_url(self):
            return reverse('post_detail_url', kwargs={'slug': self.slug})

abhiabhi94:
don't worry about the failing build, it is broken at the moment.

docs/source/setup.rst
ZendaInnocent: Thanks for the feedback

abhiabhi94 (approved):

PR #185, opened 2021-03-27
fix(#168): passing `request` object in template context

### LINKED ISSUE #168: link to redirect unauthenticated user for comment, reaction, etc. doesn't work
Screenshot (42)_LI
Screenshot (44)_LI
Screenshot (46)

### Environment
* python version 3.8
* django Version 3.1
* django-comments-dab 2.5

### Describe the bug
Note: I uninstalled the app from the packages section and put it in my directory Apps

### Steps To Reproduce
1.
2.
3.

ZendaInnocent:
Set login url next parameter to absolute url of the model object.

fixes #168

### DIFF
--- a/comment/tests/test_template_tags.py
+++ b/comment/tests/test_template_tags.py
@@ -74,6 +74,8 @@ class CommentTemplateTagsTest(BaseTemplateTagsTest):
         # no pagination
         self.assertEqual(data['comments'].count(), count)  # parent comment only
         self.assertEqual(data['login_url'], settings.LOGIN_URL)
+        # check if `request` object is passed in context template
+        self.assertEqual(data['request'], request)
 
     @patch.object(settings, 'LOGIN_URL', None)
     def test_render_comments_without_login_url(self):
--- a/comment/templatetags/comment_tags.py
+++ b/comment/templatetags/comment_tags.py
@@ -78,7 +78,8 @@ def render_comments(obj, request, oauth=False):
     context = DABContext(request, model_object=obj)
     context.update({
         'comment_form': CommentForm(request=request),
-        'oauth': oauth
+        'oauth': oauth,
+        'request': request,
     })
     return context
 


abhiabhi94 (changes requested):
Hey @ZendaInnocent, this will not fix the actual problem, it just fixes this part of the template. There are other parts that need the `request` object.

If you really want to have a stab at this, you will have to add the `request` object to the `context` dictionary here:

https://github.com/Radi85/Comment/blob/0d7bc4b99fa5573fe4bbda9ab8f32099ed8f4ffe/comment/templatetags/comment_tags.py#L74-L82
`request` is already available in this function.

Also, you will probably need to change the current tests, I think

ZendaInnocent:
Adding `request` object in `DABContext.__call__` method won't be sufficient?

I tried to add it there and seems to work fine.

abhiabhi94:
> Adding `request` object in `DABContext.__call__` method won't be sufficient?
> 
> I tried to add it there and seems to work fine.

The `DABContext` call is [used in most of the views](https://github.com/Radi85/Comment/blob/develop/comment/mixins.py#L41) when rendering a template.
When rendering a template using the method `rendering_to_string`, [a lot of metadata is passed to the template by Django](https://docs.djangoproject.com/en/dev/topics/templates/#django.template.loader.render_to_string). `request` is one of them.

So, my point is request is already available for other templates that are rendered through views. It is not available in this case, because we are using `inclusion_tag` to render the template, and it was not passed explicitly by it.

In short, adding `request` to `DABContext` seems more of an overkill to me, just adding it to the `context` dictionary used by `render_comments` template tag should be sufficient in this case.

ZendaInnocent:
Thanks for the explanations.

comment/tests/test_template_tags.py
```diff
@@ -74,6 +75,8 @@ def test_render_comments(self):
         # no pagination
         self.assertEqual(data['comments'].count(), count)  # parent comment only
         self.assertEqual(data['login_url'], settings.LOGIN_URL)
+        # check if `request` object is passed in context template
+        self.assertIsInstance(data['request'], WSGIRequest)
```
abhiabhi94, line 79: perhaps this should be:
```suggestion
        self.assertEqual(data['request'], request)
```

abhiabhi94 (approved):
Thanks!

Radi85:
@ZeroCoolHacker @abhiabhi94 
I see that @ZendaInnocent first solution is more eligible. The request object is only needed here to get the path value which is already available for us with `model_object.get_absolute_url`.

Radi85:
PS: First solution was replacing `request.path` with `model_object.get_absolute_url` in the template and it is sufficient.

abhiabhi94:
> PS: First solution was replacing request.path with model_object.get_absolute_url in the template and it is sufficient.

Actually the `request` object is also used to redirect after login for other things like :+1: , unlike, flag, subscribe etc, so adding `request` would require lesser work.

Also, `model_obj.absolute_url` would not wouldn't always be correct especially when the comment is not on the first page. 
e.g `/post?page=2`

Radi85 (approved):

Radi85:
Thanks @ZendaInnocent for your contribution!

abhiabhi94:
I think the commit message should have been something similar to the one that is the title of the issue closed at the moment. This will be potentially more helpful in generating the `changelog` and also puts the change in the perspective of the end user.
Although, I am the one to blame, I edited it as a lesser appropriate one earlier.

ZendaInnocent:
> I think the commit message should have been something similar to the one that is the title of the issue closed at the moment. This will be potentially more helpful in generating the `changelog` and also puts the change in the perspective of the end user.
> Although, I am the one to blame, I edited it as a lesser appropriate one earlier.

Amending the commit and push to the branch will work?

abhiabhi94:
> Amending the commit and push to the branch will work?

No issues, it was just a reminder for the future.

ZendaInnocent:
> > Amending the commit and push to the branch will work?
> 
> No issues, it was just a reminder for the future.

Okay, thanks.

PR #189, opened 2021-04-01
ref: Refactor tests for better debugging

### LINKED ISSUE #140: Refactor tests
### Change Type
[x ] Refactor/Enhancement
[ ] Testing
[ ] CI/CD
[ ] Other

### Proposed Changes

As of now, some unit tests are testing more than 1 unit. It would be nice to separate them. 

This would help in:

- separation of concerns.
- easier debugging.

abhiabhi94:
- almost all tests now aim to tests just one thing.
- make namesapce for DRF URLs consistent with that used for django.
- change use of raw URLs in the codebase to use url names.

closes #140

### DIFF
--- a/comment/api/urls.py
+++ b/comment/api/urls.py
@@ -3,16 +3,18 @@ from rest_framework.urlpatterns import format_suffix_patterns
 
 from comment.api import views
 
+app_name = 'comment-api'
+
 urlpatterns = [
-    path('comments/', views.CommentList.as_view(), name='comments-list'),
-    path('comments/create/', views.CommentCreate.as_view(), name='comments-create'),
-    path('comments/<int:pk>/', views.CommentDetail.as_view(), name='comment-detail'),
-    path('comments/<int:pk>/react/<str:reaction>/', views.CommentDetailForReaction.as_view(), name='comments-react'),
-    path('comments/<int:pk>/flag/', views.CommentDetailForFlag.as_view(), name='comments-flag'),
+    path('comments/', views.CommentList.as_view(), name='list'),
+    path('comments/create/', views.CommentCreate.as_view(), name='create'),
+    path('comments/<int:pk>/', views.CommentDetail.as_view(), name='detail'),
+    path('comments/<int:pk>/react/<str:reaction>/', views.CommentDetailForReaction.as_view(), name='react'),
+    path('comments/<int:pk>/flag/', views.CommentDetailForFlag.as_view(), name='flag'),
     path(
         'comments/<int:pk>/flag/state/change/',
         views.CommentDetailForFlagStateChange.as_view(),
-        name='comments-flag-state-change'
+        name='flag-state-change'
     ),
     re_path(r'^comments/confirm/(?P<key>[^/]+)/$', views.ConfirmComment.as_view(), name='confirm-comment'),
     path('comments/toggle-subscription/', views.ToggleFollowAPI.as_view(), name='toggle-subscription'),
--- a/comment/tests/test_api/test_views.py
+++ b/comment/tests/test_api/test_views.py
@@ -1,166 +1,19 @@
 from time import sleep
 from unittest.mock import patch
 
-from django.test import RequestFactory
 from django.core import signing, mail
-
+from django.urls import reverse_lazy
 from rest_framework import status
 
 from comment.conf import settings
-from comment.models import Comment, FlagInstanceManager
-from comment.messages import ContentTypeError, EmailError
+from comment.models import Comment, FlagInstanceManager, ReactionInstance
+from comment.messages import ContentTypeError, EmailError, ReactionError
 from comment.api.serializers import CommentSerializer
-from comment.api.permissions import (
-    IsOwnerOrReadOnly, FlagEnabledPermission, CanChangeFlaggedCommentState, SubscriptionEnabled,
-    CanGetSubscribers)
-from comment.api.views import CommentList
-from comment.tests.base import BaseCommentTest, timezone
+from comment.tests.base import BaseAPITest, timezone
 from comment.tests.test_utils import BaseAnonymousCommentTest
-from comment.utils import is_comment_admin, is_comment_moderator
 
 
-class APIBaseTest(BaseCommentTest):
-    @classmethod
-    def setUpTestData(cls):
-        super().setUpTestData()
-        cls.comment_1 = cls.create_comment(cls.content_object_1)
-        cls.comment_2 = cls.create_comment(cls.content_object_1)
-        cls.comment_3 = cls.create_comment(cls.content_object_1)
-        cls.comment_4 = cls.create_comment(cls.content_object_1, parent=cls.comment_1)
-        cls.reaction_1 = cls.create_reaction_instance(cls.user_1, cls.comment_1, 'like')
-
-        cls.comment_5 = cls.create_comment(cls.content_object_2)
-        cls.comment_6 = cls.create_comment(cls.content_object_2)
-        cls.comment_7 = cls.create_comment(cls.content_object_2, parent=cls.comment_5)
-        cls.comment_8 = cls.create_comment(cls.content_object_2, parent=cls.comment_5)
-        cls.reaction_2 = cls.create_reaction_instance(cls.user_1, cls.comment_5, 'dislike')
-
-    def setUp(self):
-        super().setUp()
-        self.addCleanup(patch.stopall)
-
-
-class APIPermissionsTest(APIBaseTest):
-    def setUp(self):
-        super().setUp()
-        self.owner_permission = IsOwnerOrReadOnly()
-        self.flag_enabled_permission = FlagEnabledPermission()
-        self.can_change_flagged_comment_state = CanChangeFlaggedCommentState()
-        self.subscription_enabled = SubscriptionEnabled()
-        self.get_subscribers_permission = CanGetSubscribers()
-        self.factory = RequestFactory()
-        self.view = CommentList()
-
-    @classmethod
-    def setUpTestData(cls):
-        super().setUpTestData()
-        cls.flag_data = {
-            'reason': FlagInstanceManager.reason_values[0],
-            'info': '',
-        }
-        cls.create_flag_instance(cls.user_1, cls.comment_1, **cls.flag_data)
-        cls.create_flag_instance(cls.user_2, cls.comment_1, **cls.flag_data)
-
-    def test_owner_permission(self):
-        # self.client.login(username='test-2', password='1234')
-        request = self.factory.get('/')
-        # get is in the safe methods
-        self.assertTrue(self.owner_permission.has_object_permission(request, self.view, self.comment_1))
-
-        # PUT method from different user
-        request = self.factory.put('/')
-        request.user = self.user_2
-        self.assertEqual(self.comment_1.user, self.user_1)
-        self.assertFalse(self.owner_permission.has_object_permission(request, self.view, self.comment_1))
-
-        # DELETE method from admin
-        request = self.factory.put('/')
-        request.user = self.admin
-        self.assertEqual(self.comment_1.user, self.user_1)
-        self.assertFalse(self.owner_permission.has_object_permission(request, self.view, self.comment_1))
-
-        # PUT method from same user
-        request = self.factory.put('/')
-        request.user = self.user_1
-        self.assertEqual(self.comment_1.user, self.user_1)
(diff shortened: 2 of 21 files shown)


comment/tests/test_service.py
```diff
@@ -98,18 +103,24 @@ def test_send_messages(self):
...
+
         self.email_service.send_messages(messages)
+
         self.assertNotEqual(len(mail.outbox), 100)
         self.assertTrue(self.email_service._email_thread.is_alive())
         self.email_service._email_thread.join()
+
         self.assertEqual(len(mail.outbox), 100)
 
     def test_send_confirmation_request_django(self):
```
abhiabhi94, line 117: this test along with [`test_send_confirmation_request_django`](https://github.com/Radi85/Comment/pull/189/files#diff-f2e7d9c153f774cd79acae65f97ea94af7b74f04760f3d9e37dec05bbfec5ccaR132) is flaky at the moment(at least on my machine). I'll try to investigate further about this.

Radi85: I am not sure that I understood your point.

Radi85 (changes requested):

comment/tests/test_service.py
Radi85: I am not sure that I understood your point.

abhiabhi94: if you run these tests a couple of times, they seem to fail sometimes. At least, that was the behaviour on my end.

I will try to make a video and post it, if it can't be reproduced at your level. Maybe when I can have some more peace of mind.

comment/api/urls.py
```diff
@@ -3,16 +3,18 @@
 
 from comment.api import views
 
+app_name = 'comment-api'
```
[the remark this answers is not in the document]
abhiabhi94: since the `app_name` wasn't present earlier in this module, the subsequent url names in this module were literally unusable. so this shouldn't be an issue as far as i can think.

comment/tests/test_api/test_views.py
```diff
@@ -1,180 +1,36 @@
...
             'model_name': 'post',
             'app_name': 'post',
             'model_id': 1
         }
-        self.comment_count = Comment.objects.filter_parents_by_object(self.post_1).count()
+        self.parents = Comment.objects.filter_parents_by_object(self.post_1).count()
         self.all_comments = Comment.objects.all().count()
 
+    @abc.abstractmethod
+    def get_base_url():
```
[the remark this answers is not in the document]
abhiabhi94: i agree, `abstractmethod` decorator should probably be removed.

comment/api/urls.py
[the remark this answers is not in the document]
Radi85: I think the same, just wanted to make sure.

comment/tests/test_service.py
Radi85: I feel the issue might be related to `len(mail.outbox)`. However, this never happened at my end neither in travis.

abhiabhi94: no, it is actually the key send to the email that is causing some weird issues.

In the video below, I have run the module `test_service` every 3 seconds and tried to highlight the changes in output. As you may see below, it fails sometimes.

comment-flaky-tests

Radi85: Can you share the traceback? It's not clear in the record.

abhiabhi94:
If time provides, I think we should probably have a bug fix release 2.6.1 before adding any more features. One of the bug #175 would have affected a lot of people.

Also, I think maybe we should add a `debug` flag to the javascript file and display actual errors when it is true. This will help in better error reporting and also, easier debugging hopefully.

comment/tests/test_service.py
abhiabhi94:
```sh
FAIL: test_send_confirmation_request_django (comment.tests.test_service.TestDABEmailService)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python3.8/unittest/mock.py", line 1325, in patched
    return func(*newargs, **newkeywargs)
  File "Comment/comment/tests/test_service.py", line 126, in test_send_confirmation_request_django
    self.email_contents_test(
  File "Comment/comment/tests/test_service.py", line 36, in email_contents_test
    self.assertIs(arg in contents, True, msg=f'{arg} not present in email contents')
AssertionError: False is not True : /comment/confirm/.eJxtjk0KgzAQha8i2VZlJmqhWfUmJZihBDITMclCSu_esXTZ7fve38u0Qrtx0lLqzZqlklTjzJZLjfLsvGQ5OLfSrZn5ZL0h9jGpp1KpA979yjRSaErOFAVFFiwOMA-4dGjdfHVoxxtMEy4XAAegXr9tD_FMvzFVOAdKf7Wonaj1fv--O8--P_IYOsM:1lX1Ns:c1ipHKhTIFkoAd823EnP1wQfj0Ph6wdBs0Uw1lWlqfI/ not present in email contents
```

this is when one of the test fails.

Radi85:
> If time provides, I think we should probably have a bug fix release 2.6.1 before adding any more features. One of the bug #175 would have affected a lot of people.

I am will make a new release `2.6.1` on the weekend.

> Also, I think maybe we should add a `debug` flag to the javascript file and display actual errors when it is true. This will help in better error reporting and also, easier debugging hopefully.

That would be nice. I am currently working on #190 and 403 permission will be handeled in JS so I might add the debug error as well in this PR.

comment/tests/test_service.py
abhiabhi94:
```sh

FAIL: test_send_confirmation_request_api (comment.tests.test_service.TestDABEmailService)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python3.8/unittest/mock.py", line 1325, in patched
    return func(*newargs, **newkeywargs)
  File "Comment/comment/tests/test_service.py", line 140, in test_send_confirmation_request_api
    self.email_contents_test(
  File "Comment/comment/tests/test_service.py", line 36, in email_contents_test
    self.assertIs(arg in contents, True, msg=f'{arg} not present in email contents')
AssertionError: False is not True : /api/comments/confirm/.eJxtjk0KwyAQha8ibpuEUWIhrnqTImYogjOGqItSeveOpctu3_f-XrpXPLXnnvOkY-GG3LTXR6kt8UMFLvyk0quKhWiwSSOFlMXTsLbZ3EIkXHDvQkYKd0EWrJlhnY1Txvp1825dNuvAXC8AHkC84TjuHAh_Y6JQ2TH_1ZJ0GqkP5_fdOPv-APWaOs4:1lX1RS:ex75U6gCasy3ooESdwhGVYxKYgyCfSXnjP6y7_fKZ-Y/ not present in email contents

======================================================================
FAIL: test_send_confirmation_request_django (comment.tests.test_service.TestDABEmailService)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python3.8/unittest/mock.py", line 1325, in patched
    return func(*newargs, **newkeywargs)
  File "Comment/comment/tests/test_service.py", line 126, in test_send_confirmation_request_django
    self.email_contents_test(
  File "Comment/comment/tests/test_service.py", line 36, in email_contents_test
    self.assertIs(arg in contents, True, msg=f'{arg} not present in email contents')
AssertionError: False is not True : /comment/confirm/.eJxtjk0KwyAQha8ibpuEUWIhrnqTImYogjOGqItSeveOpctu3_f-XrpXPLXnnvOkY-GG3LTXR6kt8UMFLvyk0quKhWiwSSOFlMXTsLbZ3EIkXHDvQkYKd0EWrJlhnY1Txvp1825dNuvAXC8AHkC84TjuHAh_Y6JQ2TH_1ZJ0GqkP5_fdOPv-APWaOs4:1lX1RS:ex75U6gCasy3ooESdwhGVYxKYgyCfSXnjP6y7_fKZ-Y/ not present in email contents

----------------------------------------------------------------------
Ran 15 tests in 1.188s

```

when both of them fail.

comment/tests/test_service.py
```diff
@@ -30,10 +29,11 @@ def setUp(self):
     def setUpTestData(cls):
         super().setUpTestData()
         cls.child_comment = cls.create_comment(cls.content_object_1, parent=cls.comment_1)
+        cls.email_service = DABEmailService(cls.comment_obj, cls.request)
 
     def email_contents_test(self, contents, *args):
         for arg in args:
-            self.assertTrue(arg in contents)
+            self.assertIs(arg in contents, True, msg=f'{arg} not present in email contents')
```
Radi85, line 36: What the purpose of using this form of assertion? This might be the issue since I have never encountered this fails with the old assertion `self.assertTrue(arg in contents)`.
Also where `msg=f'{arg} not present in email contents'` come from? I mean is it important info for our test?

Radi85 (changes requested):

comment/tests/test_service.py
abhiabhi94: It still doesn't work. You may check this at your end too by cloning this branch and try to run the command `watch -n 3 -d python mange.py test --keepdb comment.tests.test_service`.

> What the purpose of using this form of assertion?

This form of assertion is more explicit in this case. If you think both of them work differently, you may look up at their source [assertTrue](https://github.com/python/cpython/blob/master/Lib/unittest/case.py#L677) and [assertIs](https://github.com/python/cpython/blob/master/Lib/unittest/case.py#L1115).

> This might be the issue since I have never encountered this fails with the old assertion self.assertTrue(arg in contents)

You're right, the tests on`develop` branch are working as expected. Maybe it has got to do something with the change of `EmailService` from `setUp` method to `setUpClass` method in [`TestDABEmailService`](https://github.com/Radi85/Comment/pull/189/files#diff-f2e7d9c153f774cd79acae65f97ea94af7b74f04760f3d9e37dec05bbfec5ccaR17) or some change in the initialization of [`BaseAnonymousCommentTest`](https://github.com/Radi85/Comment/pull/189/files#diff-c218deacbe7b0c004243620d56d658670d0cb36bb5aa08a6d1bc19f6909761b7R123)or something else.

Things are pretty messed up here due to the pandemic but will try to see today if I can figure this out to distract myself.

> Also where msg=f'{arg} not present in email contents' come from? I mean is it important info for our test?

Since this test is running inside a loop, this message tells us about the argument, if and when the test fails. Otherwise, it just takes a couple of more steps to figure out exactly what caused the failure.

Radi85: Since the issue doesn't popup with the `assertTrue`, I would suggest to use it again to avoid investing more time with this issue.

abhiabhi94: does the issue not popup after the change at your end on my branch? it does on my end as i mentioned in my  previous comment https://github.com/Radi85/Comment/pull/189#discussion_r614574736, there are some other possible variables that could potentially cause this.

abhiabhi94: wow, i am pleased(i'm not sure i should be) to announce that it does actually fail on the `develop` branch as well. i'm pasting a script below that you too may use to test in your local machine. In my case, it actually failed after 62 iterations.

```bash
#!/usr/bin/bash

iteration=0
while true; do
    iteration=$((iteration + 1))
    #pytest comment/tests/test_service.py::TestDABEmailService::test_send_confirmation_request_django --pdb
    python manage.py test --keepdb comment.tests.test_service.TestDABEmailService.test_send_confirmation_request_django
    if [[ $? -ne 0 ]]; then
        echo -e "-----------\nit fucking failed in the $iteration iteration--------------------\n"
        break
    fi
    echo -e "-------------\nrunning the $iteration iteration\n-----------------\n"
done

```

Radi85: I am not sure what the purpose of running a test infinit times to see if fails or not 🤔 .

I personally cannot see where the issue is and as long as the test is passing on travis I would accept it.

I might look into this tomorrow but I don't promise 😄

abhiabhi94: > I am not sure what the purpose of running a test infinit times to see if fails or not thinking .

actually it does fail sometimes, but not always. out of the blue, it failed for me and that is how i noticed it. it is quite hard to reproduce because it passes more often than not.  i just wanted to confirm that it fails, without having to use visual feedback.

> I personally cannot see where the issue is and as long as the test is passing on travis I would accept it.

the test is flaky, that's the issue at the moment. i too am yet to figure out the exact reason for this.

abhiabhi94:
feel free to merge this, because the flakiness in the test isn't introduced by this commit. if required, it may be solved in a separate one.

Radi85 (approved):

PR #198, opened 2021-05-01
ref: Refactor tests for better debugging

abhiabhi94:
- almost all tests now aim to tests just one thing.
- make namesapce for DRF URLs consistent with that used for django.
- change use of raw URLs in the codebase to use url names.


abhiabhi94:
I have reverted the changes in the `setUp`method for the `BaseAnonymousTest` which apparently tend to make the flakiness hard to find.
Although, this means that the tests will run slower than potentially they should/could.

Radi85 (approved):

PR #199, opened 2021-05-01
feat: Add support for django 3.2

abhiabhi94:


test/settings.py
```diff
@@ -1,13 +1,13 @@
-import os
-import sys
+import os.path
```
Radi85, line 1: `os.environ` is also used in this module!

tox.ini
```diff
@@ -1,8 +1,9 @@
 [tox]
 envlist =
-    py{36, 39}-{django21, django31}
-    py{38, 39}-{djangomain}
-
+    py36-django{21, 22, 30, 31, 32}
```
Radi85, line 3: What the reason behind adding `22`, `30` and `31` versions? We agreed to reduce the test time by using the oldest and most recent version along with main. The same applies for python versions.

Radi85 (changes requested):

test/settings.py
abhiabhi94: actually `os.path` module even import `os` module behind the scenes. both of them different modules, so to be more explicit I think we can import both.

Radi85 (approved):

PR #200, opened 2021-05-03
test: Make tests run order independent

abhiabhi94:


abhiabhi94:
Initially I thought of using the `pytest` plugin [`pytest-randomly`](https://github.com/pytest-dev/pytest-randomly) to see if I could detect the flakiness in the one of the tests under the `test_services` module. 

While the flakiness still remains, I have fixed the other tests that they are not run order dependent. I later realized that the flaky test was even flaky when it was run alone, so obviously this wasn't the best idea for it.

Regarding the run-order thing, I would advise not to use raw values for `id` in the tests, because they make the tests depend upon the order in which they are run. I have fixed a few of such instances in our codebase, as of now.

Radi85 (approved):

PR #201, opened 2021-05-03
User blocker

### LINKED ISSUE #143: Block user or email from posting and reacting with comments
### Environment
* django-comments-dab version: 

### Proposed Functionality
The comment's admin or moderator can block a user/email from posting or interacting with comment.

### Use Case
To keep the comment's thread clean from naughty users 😄 and junk emails of anonymous.

### Database Changes

New table: DenyList
```python
class DenyList(models.Model):
    user = models.ForiegnKey(User, null=True, blank=True)
    email = model.EmailField(null=True, blank=True)
    blocker = models.ForiegnKey(User, null=True, blank=True)  # comment moderator
```

### LINKED ISSUE #190: Handle access permissions
### Change Type
[x] Refactor/Enhancement
[ ] Testing
[ ] CI/CD
[ ] Other

### Proposed Changes
`comment.mixins`
- Replace the default django response used in the permissions with JsonResponse to be handled properly in the JS functions.
- Move view mixins to seperate module.

Radi85:
closes #143 

- This will add the required models/managers and utils to allow comment's moderator to block users or emails of commenting or reaction with the comments.
- The view implementation along with the API are implemented in the next commit.

closes #190

- This will use status code 403 for access denied in comment mixins and handle it in JS functions.
- Implement blocking views and the API.


abhiabhi94:
can the two features be in separate pull requests? it could potentially make the diff lesser and easier to view on github.

Radi85:
> can the two features be in separate pull requests? it could potentially make the diff lesser and easier to view on github.

You can see that there is views implementation in  the second commit while it's in fact belong to the first one. It was pain to seperate them.

comment/api/permissions.py
```diff
@@ -22,6 +24,25 @@ def has_object_permission(self, request, view, obj):
...
 
 
+class UserPermittedOrReadOnly(permissions.BasePermission):
+    message = BlockedUserMSG.NOT_PERMITTED
+
+    def has_permission(self, request, view):
+        return bool(
+            request.method in permissions.SAFE_METHODS or
+            not settings.COMMENT_ALLOW_BLOCKING_USERS
+            or not BlockedUser.objects.is_user_blocked(request.user.id, request.POST.get('email'))
```
abhiabhi94, line 34: shouldn't this be a separate permission? maybe you can add this as a separate private method to `CanCreatePermission`.

comment/conf/defaults.py
```diff
@@ -35,3 +35,6 @@
 COMMENT_WRAP_CONTENT_WORDS = 30
 
 COMMENT_DEFAULT_PROFILE_PIC_LOC = '/static/img/default.png'
+
+COMMENT_ALLOW_BLOCKING_USERS = False
+COMMENT_ALLOW_MODERATOR_PERFORM_BLOCKING = False
```
abhiabhi94, line 40: maybe `COMMENT_ALLOW_MODERATOR_TO_BLOCK`?

comment/managers/blocker.py
```diff
@@ -0,0 +1,21 @@
...
+
+class BlockedUserManager(models.Manager):
+    def is_user_blocked(self, user_id=None, email=None):
+        try:
+            user_id = int(user_id)
+            return self.filter(user_id=user_id, blocked=True).exists()
+        except (ValueError, TypeError):
+            if not email:
+                return False
+            return self.filter(email=email, blocked=True).exists()
```
abhiabhi94, line 12: i feel this function can be separated in two different ones, even though function definition tells you that both parameters are not required, at least one of them is. I don't find much use of calling this without any parameters.

comment/messages.py
```diff
@@ -86,3 +88,13 @@ class FlagState:
...
     SYSTEM_NOT_ENABLED = _('Subscribe system must be enabled')
+
+
+class BlockState:
+    UNBLOCKED = _('Unblocked')
+    BLOCKED = _('Blocked')
+
+
+class BlockedUserMSG:
+    NOT_PERMITTED = _('You cannot perform this action at the moment! Contact the admin for more details')
```
abhiabhi94, line 99: i feel this message should be customizable.

comment/messages.py
```diff
@@ -86,3 +88,13 @@ class FlagState:
...
     EMAIL_REQUIRED = _('Email is required to subscribe {model_object}')
     SYSTEM_NOT_ENABLED = _('Subscribe system must be enabled')
+
+
+class BlockState:
+    UNBLOCKED = _('Unblocked')
+    BLOCKED = _('Blocked')
+
+
+class BlockedUserMSG:
```
abhiabhi94, line 98: To make this consistent with the other class names in this module
```suggestion
class BlockUserMessage:
```

comment/models/blocker.py
```diff
@@ -0,0 +1,30 @@
...
+
+    objects = BlockedUserManager()
+
+    def __str__(self):
+        return getattr(self.user, self.user.USERNAME_FIELD) if self.user else self.email
+
+
+class BlockedUserHistory(models.Model):
+    UNBLOCKED = 0
+    BLOCKED = 1
```
abhiabhi94, line 21: maybe use `enum` here?

comment/models/blocker.py
```diff
@@ -0,0 +1,30 @@
...
+
+class BlockedUser(models.Model):
+    user = models.ForeignKey(get_user_model(), on_delete=models.CASCADE, blank=True, null=True)
+    email = models.EmailField(blank=True, null=True)
+    blocked = models.BooleanField(default=True)
+
+    objects = BlockedUserManager()
+
+    def __str__(self):
+        return getattr(self.user, self.user.USERNAME_FIELD) if self.user else self.email
```
abhiabhi94, line 16: would this function be helpful here: https://github.com/Radi85/Comment/blob/develop/comment/models/comments.py#L86 ?

comment/templates/comment/block/block_modal.html
```diff
@@ -0,0 +1,39 @@
...
+                <button class="modal-close-btn" aria-label="{% trans "Close" %}">
+                    <span aria-hidden="true">&times;</span>
+                </button>
+            </div>
+            <div class="modal-body">
+                <form class="js-comment-block-form" data-url="{% url 'comment:toggle-blocking' %}">
+                {% block reason_input %}
+                    <div class="row">
+                        <div class="col-3">
+                            <label for="reason">Reason:</label>
```
abhiabhi94, line 17:
```suggestion
                            <label for="reason" class="comment-translatable"> {% trans "Reason" %} >
```
The class "comment-translatable" will allow tests to check if the content inside the element is available for translation or not, provided you are using the function `assertHttmlTranslated` when rendering this.
https://github.com/Radi85/Comment/blob/bdfdea180680b0f83730e4fdf350eed5411b539e/comment/tests/base.py#L199-L205
https://github.com/Radi85/Comment/blob/bdfdea180680b0f83730e4fdf350eed5411b539e/comment/tests/base.py#L29-L32
OR
you may also add this `label` attribute here for future purposes.
https://github.com/Radi85/Comment/blob/bdfdea180680b0f83730e4fdf350eed5411b539e/comment/tests/base.py#L26

comment/templates/comment/block/block_modal.html
```diff
@@ -0,0 +1,39 @@
+{% load i18n %}
+{% load comment_tags %}
+<div id="blockModal" class="comment-modal">
+    <div class="comment-modal-dialog" role="document">
+        <div class="comment-modal-content">
+            <div class="modal-header">
+                <h5 class="modal-title">{% block title %}{% trans "Are you sure you want to block this user?" %}{% endblock title %}</h5>
```
abhiabhi94, line 7: you can add the class `comment-translatable` here as well.

comment/templates/comment/block/block_modal.html
```diff
@@ -0,0 +1,39 @@
...
+                <button class="modal-close-btn" aria-label="{% trans "Close" %}">
```
abhiabhi94, line 8: you can add the class `comment-translatable` here as well.

abhiabhi94:
I will try to review more when I have some more time today maybe, I have only viewed close to 15 files only :weary:  as of now, which also include the rst one :laughing:  that i just skipped.

but from what i could see, this think you haven't made the change to pass an extra variable from the templates to check whether `DEBUG` is true, and show the error on the frontend for easier reporting/debugging when javascript fails.

pretty similar hack to the one that we use for `gettext` here.
https://github.com/Radi85/Comment/blob/0e22c1359257095be2f869838d10d1f3a8d48627/comment/templates/comment/static.html#L3-L10

comment/api/permissions.py
Radi85: `UserPermittedOrReadOnly` is a separate permission and it's not only used with create comment, but with all other actions.

comment/managers/blocker.py
Radi85: suggestion?

abhiabhi94: maybe, something along these lines:
```suggestion
def is_user_blocked_by_email(self, email):
    return self.filter(email=email, blocked=True).exists()
    
def is_user_blocked_by_id(self, user_id):
    return self.filter(user_id=user_id, blocked=True).exists()
```

comment/api/permissions.py
abhiabhi94: okay but i was asking you to remove the code to check if the user is blocked or not. from what i understand, blocking thing will should be used to create comment. i think blocking them to reactions etc maybe implemented at a later stage if required.

comment/managers/blocker.py
Radi85:
```python
    def is_user_blocked(self, user_id=None, email=None):
        if user_id:
            return self.is_user_blocked_by_id(user_id)
        elif email:
            return self.is_email_blocked(email)
        raise TypeError('user_id and email cannot be both None')
```

comment/messages.py
Radi85: suggestion?

comment/models/blocker.py
Radi85: I prefer the current.

comment/messages.py
abhiabhi94: I think we should probably add this to defaults with this as its default value.

comment/models/blocker.py
abhiabhi94: at other places in the codebase, we have actually used enum, so it would be better in my opinion to be consistent.

comment/messages.py
Radi85: hmmm, but this error message shows up only when the user is blocked!

abhiabhi94: so suppose i want the message to be `you are blocked to comment from this website, please container someone@domain if you think this maybe a mistake`.
 
Some might want to not include the email thing.

Someone else might just want to say `couldn't process the comment` to have some sort of abstraction.

so when passing on this message, we can do something like `settings.COMMENT_RESPONSE_FOR_BLOCKED_USER`.

comment/managers/blocker.py
abhiabhi94: why would you want to raise a typeerror, when this can be generated directly by the interpreter when using a restricted function signature, like the one i suggested above.

abhiabhi94: if you really want to do both of this in one function:
```python
def is_user_blocked(self, user_id=None, email=None):
    if not email and not user:
         raise
    if user_id:
         # do your thing
     elif:
        # do your thing
```
although python suggests the philosophy to ask for forgiveness rather than ask for permission, in this case, we won't have to do the database transaction.

although,  i'm still not convince whether writing a function signature that suggests there are two optional parameters but both of them aren't exclusively optional, is a good design strategy.

Radi85: Becasue I want to keep the logic inside the manager so that I can always use one method to check if the user blocked no matter passing the email or user id. Otherwise I need to make this checks in every place I need to check if the user is blocked. see this https://github.com/Radi85/Comment/pull/201/files#diff-150a36628526769f378f79af3ad910fb6d2e0d6c6ddf28c8cc3a20c8f5d8ba63R34 and https://github.com/Radi85/Comment/pull/201/files#diff-a3711352663d6d29db96071e9efed0a4c0c9d556402c29d56111fcc47bda266dR43

Radi85: > although, i'm still not convince whether writing a function signature that suggests there are two optional parameters but both of them aren't exclusively optional, is a good design strategy.

The original implementaion will just return False in case of not passing any params and I believe I would keep this implementation.

```python
def is_user_blocked(self, user_id=None, email=None):
        if user_id:
            return self.is_user_blocked_by_id(user_id)
        elif email:
            return self.is_email_blocked(email)
        return False
```

abhiabhi94: so, here when both `user_id` and `email` are none, you can just return `False`, without having to do the queries to the database, which will be more efficient I think.

comment/models/blocker.py
abhiabhi94: also, this may be extended to support a state where users are blocked only for a certain time period, like temporarily blocked.

comment/templates/comment/block/block_modal.html
Radi85: hmmm, this will always require us to update test translation files.

comment/messages.py
Radi85: Nice idea.
BTW, I was thinking of a way to allow devs overriding the messages.py module by creating their own module and we import in the end of our message.py.

comment/models/blocker.py
Radi85: > at other places in the codebase, we have actually used enum, so it would be better in my opinion to be consistent.

I think we have the current one 😃 https://github.com/Radi85/Comment/blob/develop/comment/models/flags.py#L15-L25

Radi85:
@abhiabhi94,
BTW, away from work. How are you doing over there with the current situation. I hope you and your family are healthy and safe.

abhiabhi94:
> BTW, away from work. How are you doing over there with the current situation. I hope you and your family are healthy and safe.

The situation is very bad in India, as you might be aware from news. Some of my family members had tested positive for coronavirus in the past month, but they are stable now. I had a freelancing/part-time job as a python instructor at a place, which is also shut as of now.

The most scare thing is the death numbers. The actual numbers that you see in news are more than 10 times in most states. The governments are hiding actual numbers, but it is sort of an open secret here. Each day, we get to hear about new deaths of nearby people or someone that we might have known.

That apart, how are you and your family there? What's the situation like there?

comment/tests/test_utils.py
```diff
@@ -98,6 +110,15 @@ def test_unauthenticated_user(self):
         self.request.user = AnonymousUser()
         self.assertIsNone(get_user_for_request(self.request))
 
+    def test_user_for_request(self):
+        request = self.factory.get('/')
+        request.user = AnonymousUser()
+        # test unauthenticated user
+        self.assertIsNone(get_user_for_request(request))
```
abhiabhi94, line 117: anonymous user is already tested in the function above: https://github.com/Radi85/Comment/pull/201/files#diff-c218deacbe7b0c004243620d56d658670d0cb36bb5aa08a6d1bc19f6909761b7L97

comment/tests/test_mixins.py
```diff
@@ -223,14 +271,78 @@ def setUp(self):
...
+        super().setUp()
+        self.toggle_blocking_url = self.get_url(reverse('comment:toggle-blocking'))
+
+    @patch.object(settings, 'COMMENT_ALLOW_BLOCKING_USERS', False)
+    def test_admin_cannot_block_users_when_system_disabled(self):
+        self.client.force_login(self.admin)
+        self.assertFalse(settings.COMMENT_ALLOW_BLOCKING_USERS)
+        data = {'comment_id': self.comment.id}
+
+        response = self.client.post(self.toggle_blocking_url, data=data, HTTP_X_REQUESTED_WITH='XMLHttpRequest')
```
abhiabhi94, line 301: I think we can override the default client with the one request this ajax header in the `setUpTestData`  function for the class`BaseCommentMixinTest`. We have done the same in `BaseCommentViewTest` here: https://github.com/Radi85/Comment/blob/develop/comment/tests/base.py#L261

abhiabhi94:
As far as i can see, i think the same permission group used for moderation is also used for blocking. in this case, we will probably also have to edit the `signals` package to make the permissions groups(in the database) when either of those settings is used.

comment/api/permissions.py
```diff
@@ -32,10 +53,7 @@ def has_permission(self, request, view):
 
 class CanChangeFlaggedCommentState(permissions.BasePermission):
     def has_permission(self, request, view):
-        return is_comment_admin(request.user) or is_comment_moderator(request.user)
-
-    def has_object_permission(self, request, view, obj):
-        return obj.is_flagged and (is_comment_admin(request.user) or is_comment_moderator(request.user))
```
abhiabhi94, line 38: why is the object level permission removed here? the thing `obj.is_flagged` should be required, no? everything else was anyway repeated.

comment/api/serializers.py
```diff
@@ -118,6 +118,8 @@ def create(self, validated_data):
         user = get_user_for_request(request)
         content = validated_data.get('content')
         email = validated_data.get('email')
+        if not user:
+            self.validate_email(email)
```
abhiabhi94, line 122: i don't think this is required. the `email` field will only be added in the `serializer` when user is unauthenticated and commenting by anonymous is allowed:  https://github.com/Radi85/Comment/pull/201/files#diff-580565eef493475112cb9daed5cc31fd758b926848bc1ba626bab78ea81acc96R102-R106

in the scenario when `email` field is added to the `serializer` fields, the `validate_email` method is already called by the parent `serializer.Serializer` class. See the documentation regarding this here: https://www.django-rest-framework.org/api-guide/serializers/#field-level-validation

comment/managers/blocker.py
```diff
@@ -0,0 +1,21 @@
...
+
+    def get_or_create_blocked_user_by_user_id(self, user_id):
+        return self.get_or_create(user_id=user_id)
+
+    def get_or_create_blocked_user_by_email(self, email):
+        try:
+            return self.get_or_create(email=email)
+        except self.model.MultipleObjectsReturned:
+            return self.filter(email=email, user=None).first(), False
```
abhiabhi94, line 21: if we want one function verifying whether the user is blocked or not, i think maybe we can do the same thing here as well. shouldn't the design pattern be consistent in a single class?

comment/mixins.py
```diff
@@ -1,162 +1,143 @@
...
+class UserPermission(BasePermission):
+    reason = BlockedUserMSG.NOT_PERMITTED
 
-    def _initialize_email_service(self, comment, request):
-        self.email_service = DABEmailService(comment, request)
+    def has_permission(self, request):
+        return bool(
+            not settings.COMMENT_ALLOW_BLOCKING_USERS or
+            not BlockedUser.objects.is_user_blocked(request.user.id, request.POST.get('email'))
+        )
```
abhiabhi94, line 44: i think this `class`is same as the one used in the `permissions` module. https://github.com/Radi85/Comment/pull/201/files#diff-150a36628526769f378f79af3ad910fb6d2e0d6c6ddf28c8cc3a20c8f5d8ba63R27-R35

I think you can probably remove one of these.

comment/mixins.py
```diff
@@ -1,162 +1,143 @@
...
-        comment.save()
-        self._send_notification_to_followers(comment, request)
-        comment.refresh_from_db()
-        return comment
+class BaseCommentPermission(UserPermission):
+    def has_permission(self, request):
+        if not request.user.is_authenticated:
+            self.reason = ErrorMessage.LOGIN_REQUIRED
+            return False
+        return super().has_permission(request)
```
abhiabhi94, line 52: i think all the work done here is again done inside its child class `BaseCreatePermission`.  https://github.com/Radi85/Comment/pull/201/files#diff-a3711352663d6d29db96071e9efed0a4c0c9d556402c29d56111fcc47bda266dR59-R64

Is this really required?

comment/tests/test_views/test_blocker.py
```diff
@@ -0,0 +1,96 @@
...
+from comment.views import BaseToggleBlockingView
+from comment.messages import BlockedUserMSG
+from comment.models import BlockedUser
+
+
+class BaseToggleBlockingViewTest(BaseCommentMixinTest):
+    def setUp(self):
+        super().setUp()
+        self.view = BaseToggleBlockingView()
+        self.toggle_blocking_url = self.get_url(reverse('comment:toggle-blocking'))
```
abhiabhi94, line 13: moving this to `setUpTestData` could/should/would potentially speed up the tests.

comment/utils.py
```diff
@@ -75,7 +75,7 @@ def has_valid_profile():
 
 
 def is_comment_admin(user):
-    if settings.COMMENT_FLAGS_ALLOWED:
+    if settings.COMMENT_FLAGS_ALLOWED or settings.COMMENT_ALLOW_BLOCKING_USERS:
```
abhiabhi94, line 78: may be we can move this to something along the lines of `_is_moderation_enabled`. it could be used in both `is_comment_admin` and `is_comment_moderator`.

comment/utils.py
```diff
@@ -84,13 +84,31 @@ def is_comment_admin(user):
...
         return user.groups.filter(name="comment_moderator").exists() or user.has_perm(
             "comment.delete_flagged_comment"
         )
     return False
 
 
+def can_moderate_flagging(user):
+    return bool(
+        settings.COMMENT_FLAGS_ALLOWED and
```
abhiabhi94, line 96: i think you can drop the `settings` check here because it is already performed in the `is_comment_moderator` function.

comment/utils.py
```diff
@@ -84,13 +84,31 @@ def is_comment_admin(user):
...
+    )
+
+
+def can_moderator_block_users(moderator):
+    return settings.COMMENT_ALLOW_MODERATOR_PERFORM_BLOCKING and is_comment_moderator(moderator)
+
+
+def can_block_user(user):
+    return bool(
+        settings.COMMENT_ALLOW_BLOCKING_USERS and
```
abhiabhi94, line 107: i think you can drop the `settings` check here because it is already performed in the `is_comment_moderator` function.

comment/models/blocker.py
```diff
@@ -0,0 +1,30 @@
...
+from django.db import models
+
+from comment.messages import BlockState
+from comment.managers import BlockedUserManager
+
+
+class BlockedUser(models.Model):
+    user = models.ForeignKey(get_user_model(), on_delete=models.CASCADE, blank=True, null=True)
+    email = models.EmailField(blank=True, null=True)
+    blocked = models.BooleanField(default=True)
```
abhiabhi94, line 11: `is_blocked` maybe? to be more explicit.

abhiabhi94: never mind, `blocked` in itself is an adjective.

comment/views/blocker.py
```diff
@@ -0,0 +1,62 @@
...
+            blocked_user=blocked_user,
+            blocker=request.user,
+            reason=reason,
+            state=int(blocked_user.blocked)
+        )
+        self.data = {
+            'blocked_user': comment.get_username(),
+            'blocked': blocked_user.blocked,
+            'urlhash': comment.urlhash
+        }
```
abhiabhi94, line 57: don't you feel this view is doing too much, some of this code could be moved to `manager`?

comment/mixins.py
```diff
@@ -1,162 +1,143 @@
...
-from comment.forms import CommentForm
-from comment.context import DABContext
-from comment.responses import DABResponseData
+from comment.messages import ErrorMessage, FlagError, FollowError, BlockedUserMSG
+from comment.responses import UTF8JsonResponse
+from comment.models import BlockedUser
 
 
 class AJAXRequiredMixin:
+    status = 403
```
abhiabhi94, line 14: if AJAX request is not allowed, should this be a bad request(400) or forbideen(403)?

Radi85:
> The situation is very bad in India, as you might be aware from news. Some of my family members had tested positive for coronavirus in the past month, but they are stable now. I had a freelancing/part-time job as a python instructor at a place, which is also shut as of now.

I hope this pandemic situation will be under control soon. Stay safe and healthy!

> That said, how are you and your family there? What's the situation like there?

We are fine and the situation seems to be under control here since the country is still locked down.

comment/models/blocker.py
Radi85: I need full email here unlike `get_username` method.

Radi85:
> but from what i could see, this think you haven't made the change to pass an extra variable from the templates to check whether `DEBUG` is true, and show the error on the frontend for easier reporting/debugging when javascript fails.

I am not sure fully understood but I added error log in JS file whenever an error thrown.

comment/tests/test_utils.py
```diff
@@ -98,6 +110,15 @@ def test_unauthenticated_user(self):
...
+        # test authenticated user
+        request.user = self.user_1
+        self.assertEqual(get_user_for_request(request), self.user_1)
```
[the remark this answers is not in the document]
Radi85: It was forgetting while rebasing.

abhiabhi94:
>I am not sure fully understood but I added error log in JS file whenever an error thrown.

so, now the error will directly be shown in the console. What i thought would be better to show these only when in developoment i.e. `DEBUG = True` inside settings. For that we will need to define a variable `debug` or some other name from the template when the project is being run in development mode, and only log errors when that variable is defined.

comment/api/permissions.py
Radi85: This method was never called since we are using `RetrieveAPIView`.
As per REST [documenation](https://www.django-rest-framework.org/api-guide/permissions/#custom-permissions) this needs object level view in order to be called.
However, this check is already done inside the view https://github.com/Radi85/Comment/blob/develop/comment/api/views.py#L113-L117

abhiabhi94: Quoting the relevant section here.

> Note: The instance-level has_object_permission method will only be called if the view-level has_permission checks have already passed. Also note that in order for the instance-level checks to run, the view code should explicitly call .check_object_permissions(request, obj). If you are using the generic views then this will be handled for you by default. (Function-based views will need to check object permissions explicitly, raising PermissionDenied on failure.)

the `RetrieveAPIView` is part of `generics`.

comment/api/serializers.py
Radi85: Hmmm as per REST [documentation](https://www.django-rest-framework.org/api-guide/serializers/#field-level-validation) this method is only triggered when the field is required. So in our case we just need to set the field to required when anonymous is used.

```python
        if not user.is_authenticated and settings.COMMENT_ALLOW_ANONYMOUS:
            self.fields['email'].required = True
```
Will do the trick.

abhiabhi94: i think by default it is required is set to `True` or whatever the model field says. also, as i see there is an `DABEmailValidator` already used in this view, where empty values which won't allow empty values. 
For better error message maybe we can move the code from `validate_email` function there?

comment/api/permissions.py
Radi85: I have tested it several time and it's indeed not called with the RetrieveAPIView, therefore I have moved the checks inside the view.

comment/api/serializers.py
Radi85: It's `False` by default, setting it to `True` will not hit the `validate_email` either but the email however validated with the default message. I am going to keep calling `validate_email` inside `create` method and I see that we no longer need the `init` method.

comment/api/permissions.py
abhiabhi94: i just saw it wasn't working, it wasn't working because `get_object` wasn't being called inside the `view` function. it was a bug on our part. also the appropriate response should be 403 in this case and not 400.

i will post a diff so that you may verify this on your end too.

abhiabhi94:
```diff
diff --git a/comment/api/views.py b/comment/api/views.py
index b0add54..0803cab 100644
--- a/comment/api/views.py
+++ b/comment/api/views.py
@@ -108,13 +108,8 @@ class CommentDetailForFlagStateChange(generics.RetrieveAPIView):
         return context
 
     def post(self, request, *args, **kwargs):
-        comment = get_object_or_404(Comment, id=kwargs.get('pk'))
+        comment = self.get_object()
         flag = Flag.objects.get_for_comment(comment)
-        if not comment.is_flagged:
-            return Response(
-                {'detail': FlagError.REJECT_UNFLAGGED_COMMENT},
-                status=status.HTTP_400_BAD_REQUEST
-            )
         state = request.data.get('state') or request.POST.get('state')
         try:
             state = flag.get_clean_state(state)

```

After applying this patch, you should run the corresponding test with the command ` python manage.py test comment.tests.test_api.test_views.APICommentDetailForFlagStateChangeTest.test_when_comment_is_not_flagged
` to verify the bug.

comment/mixins.py
Radi85: `BaseCreatePermission` has permission when anonymous system is enabled, all other acrions doesn't allow anonymity so they should be different.

comment/api/permissions.py
Radi85: Yup calling `self.get_object()` does the trick.

comment/mixins.py
Radi85: I believe it should be forbidden, 400 most likely for invalid data.

comment/api/permissions.py
abhiabhi94: i think this bug should ideally be fixed in a separate commit. if you have the time, you may otherwise i can make one.

also, for what it's worth, i don't think `RetireveAPIView` is the best choice for this view because it is modifying the object, while the former is used for mainly read-only views.

comment/utils.py
Radi85: but it doesn't for `is_comment_admin`

comment/mixins.py
abhiabhi94: also, keep in mind, this sort of change will make it backwards incompatible, because earlier, we used to return 400.

comment/utils.py
Radi85: You are right!

Radi85: I would say it safer to keep it. This check might be removed from `is_comment_moderator` in the future for some reason.

comment/mixins.py
Radi85: In fact the error were not handled in the JS earlier, it was kept to be caught in the raised error in JS function. Previously we used `HttpResponseBadRequest` which is a text response while we are parsing the response as json in the JS functions.

comment/utils.py
abhiabhi94: you might not need it(YAGNI)

Radi85:
@abhiabhi94,
Shall we merge this?

comment/api/serializers.py
```diff
@@ -99,25 +99,21 @@ class Meta:
...
+            raise serializers.ValidationError(
+                detail={'email': [EmailError.EMAIL_REQUIRED_FOR_ANONYMOUS]}, code='required'
+            )
+        return email.strip().lower()
 
     def create(self, validated_data):
         request = self.context['request']
         user = get_user_for_request(request)
         content = validated_data.get('content')
         email = validated_data.get('email')
```
abhiabhi94, line 114: `validate_email` is already called here. You can try to enter a email like `A@A.COM` with an anonymous user and see the corresponding value in the database. If the value saved inside the database is `a@a.com`, you won't need to call `validate_email` manually.

abhiabhi94, line 114: `validate_email` is already called here. You can try to enter a email like `A@A.COM` with an anonymous user and see the corresponding value in the database. If the value saved inside the database is `a@a.com`, you won't need to call `validate_email` manually.

comment/managers/blocker.py
```diff
@@ -0,0 +1,40 @@
...
+            return self.is_user_blocked_by_email(email)
+        return False
+
+    def is_user_blocked_by_id(self, user_id):
+        try:
+            return self.filter(user_id=int(user_id), blocked=True).exists()
+        except (ValueError, TypeError):
+            return False
+
+    def is_user_blocked_by_email(self, email):
```
abhiabhi94, line 22: These can be marked as internal and private.

comment/managers/blocker.py
```diff
@@ -0,0 +1,40 @@
...
+
+class BlockedUserManager(models.Manager):
+    def is_user_blocked(self, user_id=None, email=None):
+        if not settings.COMMENT_ALLOW_BLOCKING_USERS:
+            return False
+        if user_id:
+            return self.is_user_blocked_by_id(user_id)
+        elif email:
+            return self.is_user_blocked_by_email(email)
+        return False
```
abhiabhi94, line 14: you can probably save some queries to the database in some scenarios with the following:

```python
        if not user_id and not email:
            return False
        elif user_id:
            return self.is_user_blocked_by_id(user_id)
        elif email:
            return self.is_user_blocked_by_email(email)
```

comment/messages.py
```diff
@@ -86,3 +90,13 @@ class FlagState:
...
     EMAIL_REQUIRED = _('Email is required to subscribe {model_object}')
     SYSTEM_NOT_ENABLED = _('Subscribe system must be enabled')
+
+
+class BlockState:
+    UNBLOCKED = _('Unblocked')
+    BLOCKED = _('Blocked')
+
+
+class BlockUserMSG:
```
abhiabhi94, line 100: We already have a class in this module `ErrorMessage`, maybe we can be more consistent by using `BlockUserMessage`

comment/mixins.py
```diff
@@ -1,162 +1,140 @@
...
-        obj = self.get_object()
-        if not self.has_object_permission(request, obj):
-            return self.handle_no_permission()
-        return super().dispatch(request, *args, **kwargs)
+
+class CanDeleteMixin(ObjectLevelMixin, ValidatorMixin, abc.ABC):
+    def has_object_permission(self, request, obj):
+        return bool(
+            request.user == obj.user or
+            is_comment_admin(request.user) or
```
abhiabhi94, line 96: shouldn't these two be part of the `has_permission` method?

comment/tests/base.py
```diff
@@ -411,15 +427,25 @@ def setUpTestData(cls):
...
 
+    def assert_permission_denied_response(self, response, reason=None):
+        forbidden_code = 403
+        data = json.loads(response.content.decode('utf-8'))
+
+        self.assertEqual(response.status_code, forbidden_code)
+        self.assertEqual(data['status'], forbidden_code)
+        if reason:
+            self.assertEqual(data['reason'], reason)
+
```
abhiabhi94, line 448: This method is already defined in the `base class` https://github.com/Radi85/Comment/pull/201/files#diff-c2eb376f4cca55dbd53c16353ae9670ae28e0a9e50d73b2f5676ad105726fb30R307. This is just being repeated here.

comment/views/blocker.py
```diff
@@ -0,0 +1,58 @@
...
+        reason = request_data.get('reason', None)
+        if blocked_user.blocked and not reason:
+            reason = comment.content
+
+        BlockedUserHistory.objects.create(
+            blocked_user=blocked_user,
+            blocker=request.user,
+            reason=reason,
+            state=int(blocked_user.blocked)
+        )
```
abhiabhi94, line 48: This could probably be a part of the `BlockedUserHistory` manager.
```python
def create_history(*, user, comment, blocker, reason=None):
    if user.is_blocked and not reason:
         reason = comment.content
    self.create(blocked_user=user, blocker=blocker, reason=reason)
```
It can be called appropriately from the view.

Also, the setting of `state`  should ideally be inside the model class method:
```python
# models/blocker.py

def _set_state(self):
    self.state = int(self.blocked_user.blocked)

def save(self, *args, **kwargs):
    self._set_state()
    super().save(*args, **kwargs)
```

comment/managers/blocker.py
Radi85: I am not sure what you refer to but the code above won't hit the DB unless the email or user_id is provided.

docs/source/settings.rst
```diff
@@ -168,3 +168,14 @@ COMMENT_DEFAULT_PROFILE_PIC_LOC
...
+
+COMMENT_ALLOW_BLOCKING_USERS
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Enable blocking system. This gives only **admins** the right.
+
+COMMENT_ALLOW_MODERATOR_TO_BLOCK
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Allow **moderators** to perform blocking action when `COMMENT_ALLOW_BLOCKING_USERS`_ is enabled.
```
abhiabhi94, line 181: It would be informative to add the default values here. The same is done for other available configurations.

comment/api/serializers.py
Radi85: You need to call it here otherwise, missing email won't be validated.

comment/messages.py
Radi85: It should be `BlockUserError` here.

comment/mixins.py
Radi85: This is an object level permission. It doesn't make sense moving `is_comment_admin(request.user)` to `has_permission` and keeping `is_comment_moderator(request.user)` here since it's combined with obj check.

comment/tests/base.py
Radi85: You are right. My intention was to have only 1 or max 2 base test classes and we endded up with having many of them which increase the complexity. I would in the near future get rid of most of the base test classes and keep only one or 2 at most.

comment/views/blocker.py
Radi85: Agree for the manager part.

I am not sure why I need `_set_state` at all since the state can be set on the record creation only. There is no case to change an entry state at any point!

comment/api/serializers.py
abhiabhi94: This is called by default as state in the documentation about [field level validation](https://www.django-rest-framework.org/api-guide/serializers/#field-level-validation)

```sh
> git branch --show-current                                                                                                                           
develop                                                                                                                                               
git❨ develop ❩                                                                                                                                        
curious@tinker:pts/16-> /home » curious » Dev » opensourceDev » Comment (0)                                                                           
> djshell                                                                                                                                             
Python 3.8.5 (default, Jan 27 2021, 15:41:15)                                                                                                         
[GCC 9.3.0] on linux 
Type "help", "copyright", "credits" or "license" for more information.
(InteractiveConsole)
>>> from comment.api.serializers import CommentCreateSerializer as CS
>>> from django.test import RequestFactory as RF
>>> from post.models import Post
>>>
>>> post = Post.objects.last()
>>>
>>> from django.contrib.auth.models import AnonymousUser
>>>
>>> request = RF().get('/')
>>> request.user = AnonymousUser()
>>> context = {'model_obj': post, 'request': request}
>>> data = {'email': '', 'content': 'test'}
>>>
>>> from comment.conf import settings
>>> settings.COMMENT_ALLOW_ANONYMOUS = True
>>>
>>> serializer = CS(context=context, data=data)
>>> serializer.is_valid()
False
>>> serializer.data
{'email': '', 'content': 'test'}
>>> serializer.errors
{'email': [ErrorDetail(string='Email is required for posting anonymous comments.', code='required')]}
```

abhiabhi94: when we access `validated_data`, the methods for `validate_<field>` have already been called by  restframeworkk's `serializers.Serializer` class.

Radi85: Well, it seems that `serializer.is_valid()` is not called. Try this
```bash
curl -X POST -H "Content-Type: application/json" -d '{"content": "test"}' "$BASE_URL/api/comments/create/?app_name=post&model_name=post&model_id=1"
```

abhiabhi94: yes, this seems to be an unintended behavior from our side, we can probably merge this. I will try to investigate more about this later.

Radi85: We have discussed this previously. [REST docs](https://www.django-rest-framework.org/api-guide/serializers/#field-level-validation)
> Note: If your <field_name> is declared on your serializer with the parameter required=False then this validation step will not take place if the field is not included.

In our case, the email field is not required by default but it's included.

abhiabhi94 (approved):
Promise me not to make so much of changes in one pull request unless it is absolutely necessary :smile: .

comment/api/serializers.py
abhiabhi94: the validation won't take place if the field is not included, but it is included for unauthenticated users.

Radi85:
> Promise me not to make so much of changes in one pull request unless it is absolutely necessary 😄 .

I am sorry, couldn't really split it. BTW the API list view is not paginated and we need to open an issue for it.

abhiabhi94:
> I am sorry, couldn't really split it. BTW the API list view is not paginated and we need to open an issue for it.

Maybe we can keep the issue open to see if there are requirements for it to the people. If we have positive feedback, it can be implemented.

PR #202, opened 2021-05-05
fix: Response for state change on unflagged comments

abhiabhi94:
- also change use of `generics.RetrieveAPIView` to `generics.UpdateAPIView` in API views as these views were used to modify an instance.

- both of these changes only apply to the usage of REST API through DRF.

### DIFF
--- a/comment/api/views.py
+++ b/comment/api/views.py
@@ -46,7 +46,7 @@ class CommentDetail(generics.RetrieveUpdateDestroyAPIView):
     permission_classes = (permissions.IsAuthenticatedOrReadOnly, IsOwnerOrReadOnly, UserPermittedOrReadOnly)
 
 
-class CommentDetailForReaction(generics.RetrieveAPIView):
+class CommentDetailForReaction(generics.UpdateAPIView):
     queryset = Comment.objects.all()
     serializer_class = CommentSerializer
     permission_classes = (permissions.IsAuthenticatedOrReadOnly, UserPermittedOrReadOnly)
@@ -74,7 +74,7 @@ class CommentDetailForReaction(generics.RetrieveAPIView):
         return Response(serializer.data, status=status.HTTP_200_OK)
 
 
-class CommentDetailForFlag(generics.RetrieveAPIView):
+class CommentDetailForFlag(generics.UpdateAPIView):
     queryset = Comment.objects.all()
     serializer_class = CommentSerializer
     permission_classes = (permissions.IsAuthenticatedOrReadOnly, FlagEnabledPermission, UserPermittedOrReadOnly)
@@ -98,7 +98,7 @@ class CommentDetailForFlag(generics.RetrieveAPIView):
         return Response(serializer.data, status=status.HTTP_200_OK)
 
 
-class CommentDetailForFlagStateChange(generics.RetrieveAPIView):
+class CommentDetailForFlagStateChange(generics.UpdateAPIView):
     queryset = Comment.objects.all()
     serializer_class = CommentSerializer
     permission_classes = (CanChangeFlaggedCommentState, UserPermittedOrReadOnly)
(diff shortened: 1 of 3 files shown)


abhiabhi94:
Actually, I think it isn't a great option to test the `permission` module individually, since they are always going to be used with the `views`, I think it would be a better thing to take some dummy views in the `test_permissions.py` module and then test their response and status code.

This idea is basically copied from the way `djangorestframework` [tests their permissions](https://github.com/encode/django-rest-framework/blob/master/tests/test_permissions.py).

Radi85:
> Actually, I think it isn't a great option to test the `permission` module individually, since they are always going to be used with the `views`, I think it would be a better thing to take some dummy views in the `test_permissions.py` module and then test their response and status code.

The permissions are tested along with the views as well so we don't need to repeat the process.

abhiabhi94:
> The permissions are tested along with the views as well so we don't need to repeat the process.

Yes, that is the point. We wouldn't need to test them again with views.

comment/api/views.py
```diff
@@ -45,7 +45,7 @@ class CommentDetail(generics.RetrieveUpdateDestroyAPIView):
     permission_classes = (permissions.IsAuthenticatedOrReadOnly, IsOwnerOrReadOnly)
 
 
-class CommentDetailForReaction(generics.RetrieveAPIView):
+class CommentDetailForReaction(generics.UpdateAPIView):
```
Radi85, line 48: There won't be any change on the comment itself, why would we need UpdateAPIView?

Radi85 (changes requested):

comment/api/views.py
abhiabhi94: yes but the `flag` object is part of the comment instance. so we are actually updating an instance here.

abhiabhi94: `RetrieveAPIView` is anyway used for read-only situations.

Radi85: > `RetrieveAPIView` is anyway used for read-only situations.

This is exactly what is intended to be when updating one of the comment components (flag, reaction ...), I don't want the comment to be accidentally updated.

abhiabhi94: what is the object under consideration for the view? it's a comment, a part of which is getting modified.

for what it's worth, the name of the view doesn't restrict anything, it just gives you some handlers for some cases. In our case, we are anyway overriding the `post` method and manually returning the deserialized instance.

Radi85: The view name is matter here, this will give the user hint that the comment itself should not be updated within this view.

Radi85: Are you insisting on using `UpdateAPIView`? If yes, please test the affected endpoints by passing the comment content in the body and if you get the expected result, we can then merge it.

abhiabhi94: don't you think the current tests don't cover this?

Radi85: I don't think that unit tests cover 100% of the cases even if the coverage ratio is 100%. E.g. The unit test was covering create comment in the serializer but we found that `validate_email` method was not called as expected. Therefore integration test is needed for some cases.

abhiabhi94: okay, i have added a test to demonstrate this.

comment/tests/test_api/test_views.py
```diff
@@ -320,6 +320,18 @@ def get_base_url(self, reaction, c_id=None):
             c_id = self.comment.id
         return reverse_lazy('comment-api:react', args=[c_id, reaction])
 
+    def test_cannot_update_comment_content(self):
+        comment = self.comment
+        original_content = comment.content
+        data = {'content': 'test updation during reactions'}
+
+        response = self.client.post(self.get_base_url(self.like), data=data, content_type='application/json')
```
Radi85, line 328: What is the response data? Why is the statusCode 200? Shouldn't it be 400?

Radi85 (changes requested):

comment/tests/test_api/test_views.py
abhiabhi94: the status code is 200 because as designed currently we have made the content field read only. So, the changes are ignored but no errors are thrown.

the response data is the one returned for a successful reaction on a comment.

abhiabhi94: i have also added the test for response data as well.

Radi85 (approved):

PR #209, opened 2021-05-29
test: Speed up tests by more than 500%

abhiabhi94:
- use django's `SimpleTestCase` wherever database access is not required.
- use a separate module for testing that is lesser secure but faster.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,9 @@ env:
 install: pip install tox-travis coveralls
 script:
   - tox
-after_success: coveralls
+after_success:
+  - coverage combine
+  - coveralls
 
 jobs:
   allow_failures:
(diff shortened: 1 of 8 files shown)


abhiabhi94:
For some unscientific comparisons, these are the results when run on my machine:

Before:
```
Ran 395 tests in 68.571s
```

After:
```
Ran 395 tests in 10.824s
```

abhiabhi94:
I have created a separate `settings` module for tests because I thought you @Radi85, use the current one for live demo. In case you don't, I can make the changes directly to the original `settings` module.

Radi85:
> I have created a separate `settings` module for tests because I thought you @Radi85, use the current one for live demo. In case you don't, I can make the changes directly to the original `settings` module.

Yes I am using it in the live demo.

Radi85 (approved):
Well done!

PR #211, opened 2021-06-02
fix(#210): Reduce chances of XSS injections

### LINKED ISSUE #210: xss injection
### Environment
* python version 3.8.4
* django Version 3.2.2
* django-comments-dab 2.6

### Describe the bug
your form has no protection against xss injections
Sorry in advance, but I checked it out on your live demo
don't worry i deleted my comment

### Steps To Reproduce
1. publish this code in form for comment
<script>alert('Test alert');</script>

### Screenshots
image

abhiabhi94:
- on the content field.

fixes #210

### DIFF
--- a/comment/templates/comment/comments/content.html
+++ b/comment/templates/comment/comments/content.html
@@ -4,10 +4,10 @@
     {% include "comment/comments/urlhash.html" with urlhash=urlhash %}
     {% endblock comment_urlhash %}
     {% if text_2 %}
-        {{ text_1|safe }}
-        <span class="d-none">{{ text_2|safe }}</span>
+        {{ text_1 }}
+        <span class="d-none">{{ text_2 }}</span>
         <button class="js-read-more-btn {% block read_more_cls %}btn btn-link btn-xs read-more{% endblock read_more_cls %}">{% trans "read more" %}...</button>
     {% else %}
-        {{ text_1|safe }}
+        {{ text_1 }}
     {% endif %}
 </p>
--- a/comment/templatetags/comment_tags.py
+++ b/comment/templatetags/comment_tags.py
@@ -1,6 +1,8 @@
 import re
 
 from django import template
+from django.utils.html import conditional_escape
+from django.utils.safestring import mark_safe
 
 from comment.models import ReactionInstance, FlagInstance, Follower, BlockedUser
 from comment.forms import CommentForm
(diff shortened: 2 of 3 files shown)


comment/templates/comment/comments/content.html
```diff
@@ -4,10 +4,10 @@
     {% include "comment/comments/urlhash.html" with urlhash=urlhash %}
     {% endblock comment_urlhash %}
     {% if text_2 %}
-        {{ text_1|safe }}
```
Radi85, line 7: I believe this will break this #169

Radi85 (changes requested):

comment/templates/comment/comments/content.html
abhiabhi94: yes, you are right. I have made changes to the `render_content` template tag to first escape the original content and then marked it as safe after our change.

Radi85 (dismissed):

Radi85:
Great work!
@abhiabhi94 Could you please rebase your PR?

abhiabhi94:
Hey @Radi85, this status reporting in taking too long on travis-ci. It maybe due to an issue at either end. Do you consider switching to GitHub actions? A lot of open-source repositories have migrated to it, and I too have had satisfactory experience.

Although, It maybe a backroom strategy by Microsoft to have this migration.

Radi85:
> Hey @Radi85, this status reporting in taking too long on travis-ci. It maybe due to an issue at either end. Do you consider switching to GitHub actions? A lot of open-source repositories have migrated to it, and I too have had satisfactory experience.

I have no never used actions but I would not mind to migrate since we are facing issues with Travis. I will have a look at it as soon as I have a time.

> Although, It maybe a backroom strategy by Microsoft to have this migration.

Microsoft 😆

Radi85 (approved):

abhiabhi94:
> I have no never used actions but I would not mind to migrate since we are facing issues with Travis. I will have a look at it as soon as I have a time.

I will try to make a patch for this one soon.

abhiabhi94:
One more thing, you'll have to add the password token to the secrets for automated release to PyPI with actions. I would suggest using a manual release as there have been instances in the past when secrets have been leaked(also I don't like microsoft).

I can handle the release process, if you don't have the time. In case, you have any other suggestions, they are welcome.

Issue #153, opened 2021-01-20
Add support for rendering comment in markdown format


abhiabhi94:
### Proposed Functionality
I think we can probably use a template tag with the help of a third-party package to render comment content in `markdown` format. Comment truncation should be disabled when this functionality is enabled.

### Use Case
Adds highlighting of codes and other stuff.

### Database Changes
None.

Radi85:
@abhiabhi94 
Which release do you think that we can include this?

Radi85:
I believe we need to increase the height of the textarea when on focuse and this should be applicable when this feature is enabled. I would use 4 rows on focus.
Do you think we can use helper widgets with the textarea?

abhiabhi94:
> Which release do you think that we can include this?

```python
if release('2.6').planned_date > `this week end`:
    set_release('2.7')
else:
    set_release('2.6')
```

> Do you think we can use helper widgets with the textarea?

Yes, should be able to. As of now, I think this should be sufficient for our use case. I can only comment more when I start working on this.

Radi85:
@abhiabhi94, How's the progress here? I will postpone the next release until this is done.

abhiabhi94:
> @abhiabhi94, How's the progress here? I will postpone the next release until this is done

As mentioned in https://github.com/Radi85/Comment/pull/212#issuecomment-867758343, it was being blocked. I will try to start the work for this as soon as I can now.

As a side note, I will think we can probably shift this feature to the next release if possible, since there was a recent security bug fix #211. It would be great if that can be released, prior to this. This is obviously considering if everything else(i.e. time constraint ,etc,) permit.

Radi85:
> As a side note, I will think we can probably shift this feature to the next release if possible, since there was a recent security bug fix #211. It would be great if that can be released, prior to this. This is obviously considering if everything else(i.e. time constraint ,etc,) permit.

Shall we move this to `2.8.0`?

abhiabhi94:
yes, in my opinion that would be better.

PR #212, opened 2021-06-24
chore(#206): Move metadata from setup.py to setup.cfg

### LINKED ISSUE #206: Provide wheels alongside source distribution on PyPi
### Change Type
[ ] Refactor/Enhancement
[ ] Testing
[ x] CI/CD
[ ] Other

### Proposed Changes

As of now, we only provide source distribution on PyPi. See https://pypi.org/simple/django-comments-dab/

We should look to provide `wheels` alongside source distribution. The wheel distribution will make it faster for users to install the package.

While we are at it, I think we should consider moving the metadata from  `setup.py` to `setup.cfg`. It is a more declarative format of metadata, and `setup.py` is more of a legacy format, mostly used to make editable installations work.

You can read more about that at: https://packaging.python.org/tutorials/packaging-projects/#configuring-metadata

abhiabhi94:
- also add `pyproject.toml` to specify build time dependencies.
- stop distributing tests and `pyc` files to reduce package size.

resolves #206


abhiabhi94:
this is blocking #153 as the approach for optional installation  would be built on the top of it.

Radi85 (approved):
@abhiabhi94 thanks for your efforts.

abhiabhi94:
@Radi85 Thanks, for checking.

PR #213, opened 2021-07-09
fix: Namespace pollution caused by star imports

abhiabhi94:
- use explicit imports and `__all__` to declare allowed list of public objects.
- some imports are still provided for backward compatibility, they may be removed in a major version bump.
- also, use correct packages for import of managers in tests.

### DIFF
--- a/comment/models/__init__.py
+++ b/comment/models/__init__.py
@@ -1,6 +1,30 @@
-# flake8: noqa
-from comment.models.comments import *
-from comment.models.reactions import *
-from comment.models.flags import *
-from comment.models.followers import *
-from comment.models.blocker import *
+from comment.models.comments import Comment
+from comment.models.reactions import Reaction, ReactionInstance
+from comment.models.flags import Flag, FlagInstance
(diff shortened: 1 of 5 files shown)


comment/models/__init__.py
```diff
@@ -1,6 +1,30 @@
...
+__all__ = (
+    'Comment',
+    'Reaction',
+    'ReactionInstance',
+    'Flag',
+    'FlagInstance',
+    'Follower',
+    'BlockedUser',
+    'BlockedUserHistory',
+    # TODO: managers are given here due to the earlier namespace pollutin caused by star imports,
```
Radi85, line 20: Can we open an issue for this?

Radi85 (dismissed):

comment/models/__init__.py
abhiabhi94: #217 aims to track this.

Radi85 (approved):

abhiabhi94:
Thanks Radico, for checking.

PR #222, opened 2021-08-18
feat(#153): Add support for rendering content in markdown format

abhiabhi94:
resolves #153

PR #226, opened 2021-08-18
chore(#225): Run tests for all supported python and django versions

### LINKED ISSUE #225: Run tests for all supported python and django versions
### Change Type
[ ] Refactor/Enhancement
[ x] Testing
[ ] CI/CD
[ ] Other

### Proposed Changes

As of now, we run tests only for the latest and oldest supported python and Django versions. This might not be ideal in every case, because something that might work on Django `3.2` doesn't necessarily work on `3.1` because `3.2` is LTS(long term support) so patches are still back ported to it. Whereas `3.1` is in freeze mode and hence it does no longer receive back ports.

Similarly, for `2.1` and `2.2`(LTS).

In order to not get caught in a situation where a test failure wouldn't be caught by our test suite, I think we should run them on all supported python and Django versions. I understand that we had made this choice to run only on some versions earlier to reduce the testing time. A lot on this front has been taken care now that tests run in fairly less amount of time and tests for different versions can be run parallelly(if this is not true on `travis`, maybe we can switch to GitHub actions).

Also, i found that we don't run the tests on Django `3.2` on travis, this was probably missed in https://github.com/Radi85/Comment/commit/e171d8f0ca32d79f31a1b7bb3ba0fe0cf5b526f0. This too can be handled in this.

This is also blocking #220

abhiabhi94:
fixes #225

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,10 @@ python:
   - 3.9
 env:
   - DJANGO=2.1
+  - DJANGO=2.2
+  - DJANGO=3.0
   - DJANGO=3.1
+  - DJANGO=3.2
   - DJANGO=main
 
 install: pip install tox-travis coveralls
--- a/tox.ini
+++ b/tox.ini
@@ -1,9 +1,9 @@
 [tox]
 envlist =
-    py36-django{21, 32}
-    py37-django{21, 32}
-    py38-django{21, 32, main}
-    py39-django{21, 32, main}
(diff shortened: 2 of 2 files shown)


abhiabhi94:
wow, the whole test suite ran in just 3 minutes, while the earlier ones with lesser number of jobs like #222 took 5 minutes 🤔

Radi85 (approved):

abhiabhi94:
Oh! I missed python `3.7` in Travis! 😠

PR #220, opened 2021-08-17
chore: release 2.7.0

Radi85:
closes #218

PR #228, opened 2021-08-19
chore: Migrate to github actions for CI

### LINKED ISSUE #223: Install `wheels` in Continuous Integration(CI) environments
### Change Type
[ ] Refactor/Enhancement
[ ] Testing
[x] CI/CD
[ ] Other

### Proposed Changes

Installing `wheels` in Continuous Integration(CI) environments will help reduce installation time for dependencies that distribute wheel.

abhiabhi94:
- add separate environment in tox for linters and docs.
- switch to codecov for coverage since coveralls doesn't play well with github-actions.

Thanks travis for their service till now.

closes actions/toolkit#223.


abhiabhi94:
Hey @Radi85, you may have to set up the repository on codecov, although their bot is still commenting so i guess it may not be required.

I haven't added anything related to release as of now, please let me know your thoughts regarding the potential approaches as stated in https://github.com/Radi85/Comment/pull/220#issuecomment-902274135. any new ideas are welcome.

Radi85:
Hey @abhiabhi94, well done. Thanks a lot for your time.
I have updated the settings, removed travis things and so on. I also synchronize codecov with the repo.
I would like to add the release at least for testing pypi and the main one will be kept to manually for now. Can you please add the test pypi release to the CI?

abhiabhi94:
> I have updated the settings, removed travis things and so on. I also synchronised codecov with the repo.

Thank you 🎉 for the updates.

> I would like to add the release at least for testing pypi and the main one will be kept to manually for now. Can you please add the test pypi release to the CI?

I have updated the patch to include the [release workflow](https://github.com/Radi85/Comment/pull/228/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34). The aim is to release to test PyPI servers when there is tag pushed on the `master` branch. This was what I could gather from our previous `travis` workflow. Please inform if this doesn't seem suffice for our use-case or anything else needs attention. Happy to make the changes.

You probably would have to setup the credentials on Github, in case you haven't done it yet. This [guide](https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#saving-credentials-on-github) could be helpful.

.github/workflows/release.yml
```diff
@@ -0,0 +1,41 @@
+name: Release
```
Radi85, line 1:
```suggestion
name: TestRelease
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,41 @@
...
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+    tags:
+      - '**'  # allow every tag including hierarchical tags like v1.0/beta
```
Radi85, line 9: `staging` branch should be released to TestPypi
```suggestion
    branches:
      - staging
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,41 @@
...
+
+    - name: Build wheel and source distribution
+      run: |
+          python setup.py sdist bdist_wheel
+
+    - name: Publish distribution 📦 to Test PyPI
+      if: startsWith(github.ref, 'refs/tags')
+      uses: pypa/gh-action-pypi-publish@master
+      with:
+        user: __token__
```
Radi85, line 36: do we need to prived the user here?

.github/workflows/release.yml
```diff
@@ -0,0 +1,41 @@
...
+  push:
+    branches:
+      - master
+    tags:
+      - '**'  # allow every tag including hierarchical tags like v1.0/beta
+
+jobs:
+  build-n-publish:
+    name: Build and publish Python 🐍 distributions 📦 to TestPyPI
+    runs-on: ubuntu-latest
```
Radi85, line 14: Can we allow this job on failure?
```suggestion
    runs-on: ubuntu-latest
    continue-on-error: true
```

Radi85 (changes requested):

.github/workflows/release.yml
abhiabhi94: this file can be used to setup an automated release to PyPI as well, although it is not being used as now. If you see the [step-name](https://github.com/Radi85/Comment/pull/228/files/087cea7a81c1642be483141f25d00304c2d04df3#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R32), it does state more explicit information. Do you still want to change this?

.github/workflows/release.yml
abhiabhi94: `continue-on-error` is step-specific. At job level, [there's still an issue open in GitHub](https://github.com/actions/toolkit/issues/399).
For the time being, you may not set this as required as a status check, maybe?

.github/workflows/release.yml
Radi85: We agreed that test release only will be used in action pipeline so it makes more sense for me to use `TestRelease`. This can be updated in the future if we go for the main release.

.github/workflows/release.yml
Radi85: I understood from the github opened issue that `continue-on-error` is used for allowing the job to fail. However it's necessary to allow failure for TestRelease job since the release will definitely fail on the second try when using the same release version and to avoid changing the version back and forth while preparing for the main release.

abhiabhi94: i'm just quoting the reply of one of the collaborators of the `actions/toolkit` repository on the opened issue: https://github.com/actions/toolkit/issues/399#issuecomment-607450398(removed the mention to the original poster here)
> Hey, Thanks for the feature request!

>We do support marking a step to allow failure via continue-on-error.
We also support marking specific checks as required
That will allow you to merge pr's without a particular job succeeding.

> I think the latter should solve your issue at the job level, is there a reason that doesn't well for you?

So, basically, from what I understand here we can't mark the `continue-on-error` check at job-level, for that we should not make the status for this check as required. This will allow failures from this job to be allowed.

Radi85: So let's make as not required.

abhiabhi94:
hey @Radi85, is their anything other change that is required in this patch or is this now good to go?

Radi85 (approved):

Radi85:
> hey @Radi85, is their any other change that is required in this patch or is this now good to go?

So far I think we are good to go. Let's see how things go and later on we decide whether adding the main release or just keeping it manually.

@abhiabhi94 thanks for the good work.

PR #229, opened 2021-08-29
fix: docs issue

Radi85:


README.rst
```diff
@@ -6,11 +6,11 @@
     :target: https://pypi.python.org/pypi/django-comments-dab/
     :alt: django
 
-.. image:: https://codecov.io/gh/radi85/Comment/branch/main/graph/badge.svg?token=JBorE9i0De
+.. image:: https://codecov.io/gh/radi85/Comment/branch/master/graph/badge.svg?token=JBorE9i0De
```
abhiabhi94, line 9: Shouldn't this be for the `develop` branch?

```suggestion
.. image:: https://codecov.io/gh/radi85/Comment/branch/develop/graph/badge.svg?token=JBorE9i0De
```

README.rst
```diff
@@ -6,11 +6,11 @@
...
     :target: https://codecov.io/gh/radi85/Comment
     :alt: Coverage
 
-.. image:: https://github.com/radi85/Comment/actions/workflows/tests.yml/badge.svg?branch=main
+.. image:: https://github.com/radi85/Comment/actions/workflows/tests.yml/badge.svg?branch=master
```
abhiabhi94, line 13: ditto

README.rst
Radi85: I think this should be the stable version which is master, right?

abhiabhi94: i think `readthedocs` give you the option to go to enable different versions and set `stable` as the default version. keeping this to `develop` would help us see the documentation builds instantly similar to test badges.

abhiabhi94 (approved):
Thanks, @Radi85 for the patch.

PR #231, opened 2021-09-04
fix(#230): Fix closing of anonymous create comment modal

### LINKED ISSUE #230: Javascript error on cancel of the block user modal.
### Environment
* python version 3.6
* django Version 3.2
* django-comments-dab 2.7.0

### Describe the bug
Closing the "block user" modal raises a javascript error and the modal fails to close.

```
15:05:52.214 comment.js:41 Uncaught TypeError: Cannot read property 'classList' of null
    at hideModal (comment.js:41)
    at HTMLDocument.<anonymous> (comment.js:790)
```

### Steps To Reproduce
1. post a comment
2. login as admin
3. click the lock icon to block a user, and cancel the block modal

### Expected behavior
Cancel the action and close the modal.

abhiabhi94:
- the issue occured if and when the modal wasn't active and the hideModal was called to close it.

fixes #230

### DIFF
--- a/comment/static/js/comment.js
+++ b/comment/static/js/comment.js
@@ -31,6 +31,7 @@ document.addEventListener('DOMContentLoaded', () => {
     };
 
     let showModal = modalElement => {
+        if (!modalElement) return;
         modalElement.style.display = 'block';
         setTimeout(() => {
             modalElement.classList.add('show-modal');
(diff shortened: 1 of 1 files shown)


abhiabhi94:
hey @Radi85, you may remove the `docs` check from required, since(as of now) it is written to build only when there are changes in the `docs` directory, just to stop unnecessary builds.

Radi85 (changes requested):
We should check the `modelElement` if exists on [`showModal` and `hideModal`](https://github.com/Radi85/Comment/blob/develop/comment/static/js/comment.js#L33-L43) as well.

```python
let showModal = modalElement => {
        if (!modalElement) return;
        modalElement.style.display = 'block';
        setTimeout(() => {
            modalElement.classList.add('show-modal');
        }, 20);
    };

    let hideModal = modalElement => {
        if (!modalElement) return;
        modalElement.classList.remove('show-modal');
        modalElement.style.display = 'none';
    };
```

abhiabhi94:
Hey @Radi85, thanks for taking the time for checking this. Yes, these change seems legit to me.

Radi85 (approved):
@abhiabhi94 thanks for the fix.

PR #232, opened 2021-09-06
chore: release 2.7.1

Radi85:


abhiabhi94 (approved):
thanks for the patch.

Radi85:
@abhiabhi94 `Publish distribution 📦 to Test PyPI` job was not triggered any idea?

abhiabhi94:
@Radi85 was this a tagged commit? as of now, i can't see the tag on the branch `staging` although the build logs suggest that binaries were created to distribute, but the publishing task didn't execute. weird i guess 😕

abhiabhi94:
one more thing, i merged the commit since i thought it to be a waste of time if the patch was approved by one of us, and waited for the other to merge. please let me know, if doesn't seem right.

Radi85:
> @Radi85 was this a tagged commit? as of now, i can't see the tag on the branch `staging` although the build logs suggest that binaries were created to distribute, but the publishing task didn't execute. weird i guess 😕

The job should be trigered after pushing staging branch and the tag is not required here! 🤔 
BTW in the previous release the above was applied am not sure why it wasn't triggered here.

Radi85:
> one more thing, i merged the patch since i thought it to be a waste of time if it was approved by one of us, and just waited for merging. please let me know, if doesn't seem right.

I am not sure how the right process could be. However, I would prefer to handle the whole merging processe for the **releases** and don't mind when one of us merge the fueatures/bugfixes.

abhiabhi94:
> The job should be trigered after pushing staging branch and the tag is not required here! 🤔
BTW in the previous release the above was applied am not sure why it wasn't triggered here.

no, it should only be triggered on tags, see: https://github.com/Radi85/Comment/blob/develop/.github/workflows/release.yml#L31

Radi85:
> no, it should only be triggered on tags, see: https://github.com/Radi85/Comment/blob/develop/.github/workflows/release.yml#L31

Then let's remove the tag condition. I am not sure how did it work in the previous release without tag.

PR #244, opened 2022-04-28
Using `exists` instead of `count` is faster.

nshaibu:


abhiabhi94 (approved):
thanks for the contribution!

PR #247, opened 2022-06-20
chore: add main release workflow

Radi85:


.github/workflows/release.yml
```diff
@@ -1,14 +1,13 @@
-name: Test Release
+name: Release
 
 on:
-  pull_request:
   push:
-    branches:
-      - staging
+    tags:
```
abhiabhi94, line 5: add the check for `branches` here as well? only push from `develop`? otherwise all tags even if accidentally pushed to some other branch may trigger a release.

.github/workflows/test-release.yml
```diff
@@ -0,0 +1,38 @@
...
+    - name: Ensure latest setuptools
+      run: |
+          python -m pip install --upgrade pip setuptools wheel
+
+    - name: Build wheel and source distribution
+      run: |
+          python setup.py sdist bdist_wheel
+
+    - name: Publish distribution 📦 to Test PyPI
+      uses: pypa/gh-action-pypi-publish@master
```
abhiabhi94, line 31: earlier this used to be done only with tagged commits. now all commits on staging branch will get pushed? any particular reason for this change?

abhiabhi94 (changes requested):

.github/workflows/test-release.yml
abhiabhi94: my bad, you're correct.

abhiabhi94 (approved):

abhiabhi94:
thanks for the contribution, Radico.

PR #250, opened 2022-07-19
translate the persian language (fa)

AmirAref:


AmirAref:
to make new pr include new commit

Issue #259, opened 2022-11-03
docker-compose error on building sample project.


NaGaii1994:
### Environment
* os: ubuntu 20.04LTS
* docker version: Docker version 20.10.12, build 20.10.12-0ubuntu2~20.04.1

### Describe the bug
Hello, thanks for this useful software !
I have erorr on building sample project on docker.
```
Building wheels for collected packages: backports.zoneinfo
  Building wheel for backports.zoneinfo (pyproject.toml): started
  Building wheel for backports.zoneinfo (pyproject.toml): finished with status 'error'
  error: subprocess-exited-with-error
  
  × Building wheel for backports.zoneinfo (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [37 lines of output]
      running bdist_wheel
      running build
      running build_py
      creating build
      creating build/lib.linux-x86_64-cpython-38
      creating build/lib.linux-x86_64-cpython-38/backports
      copying src/backports/__init__.py -> build/lib.linux-x86_64-cpython-38/backports
      creating build/lib.linux-x86_64-cpython-38/backports/zoneinfo
      copying src/backports/zoneinfo/_version.py -> build/lib.linux-x86_64-cpython-38/backports/zoneinfo
      copying src/backports/zoneinfo/_common.py -> build/lib.linux-x86_64-cpython-38/backports/zoneinfo
      copying src/backports/zoneinfo/_tzpath.py -> build/lib.linux-x86_64-cpython-38/backports/zoneinfo
      copying src/backports/zoneinfo/__init__.py -> build/lib.linux-x86_64-cpython-38/backports/zoneinfo
      copying src/backports/zoneinfo/_zoneinfo.py -> build/lib.linux-x86_64-cpython-38/backports/zoneinfo
      running egg_info
      writing src/backports.zoneinfo.egg-info/PKG-INFO
      writing dependency_links to src/backports.zoneinfo.egg-info/dependency_links.txt
      writing requirements to src/backports.zoneinfo.egg-info/requires.txt
      writing top-level names to src/backports.zoneinfo.egg-info/top_level.txt
      reading manifest file 'src/backports.zoneinfo.egg-info/SOURCES.txt'
      reading manifest template 'MANIFEST.in'
      /tmp/pip-build-env-lvfy0dt6/overlay/lib/python3.8/site-packages/setuptools/config/setupcfg.py:508: SetuptoolsDeprecationWarning: The license_file parameter is deprecated, use license_files instead.
[...]

PR #260, opened 2022-11-03
fix(#259): docker-compose error on building sample project.

NaGaii1994:
- Change base image from alpine to slim-bullseye because using alpine can make python docker slower.
- Update postgres image from 9.6 to 14.
- Fix bug that django container connect database before database is ready.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,4 +16,5 @@ COPY . /code/
 WORKDIR /code/
 
 RUN pip install -r /code/test/example/requirements.txt \
+    && pip install psycopg2 \
     && python -m pip install django-comments-dab[markdown]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,21 +1,19 @@
-FROM python:3.8-alpine
+FROM python:3.10-slim-bullseye
 
-RUN apk update \
-    && apk add --virtual build-deps gcc python3-dev musl-dev \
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -1,21 +1,19 @@
-FROM python:3.8-alpine
+FROM python:3.10-slim-bullseye
```
abhiabhi94, line 1: can we use 3.11, since it's out now?

test/example/requirements.txt
```diff
@@ -10,3 +10,5 @@ cssselect
 coverage
 flake8
 tox
+# used for postgresql connection
+psycopg2
```
abhiabhi94: can we still keep this in docker file? if it's here, not everyone that uses the docker file will have to install postgres. people can use this locally on sqlite without docker.

abhiabhi94 (commented):
hey @NaGaii1994 thanks for taking the time to contribute. left some comments, rest of the patch looks fine to me.

Dockerfile
NaGaii1994: yes, I confirmed that unit test is passed.  But should I not change that of version?

test/example/requirements.txt
NaGaii1994: ok. I did that on another commit. Please check it.

Issue #263, opened 2023-12-15
multiple comment fields in one page


ChathuraGH:
### Proposed Functionality
1) multiple comment fields in one page (currently only one is possible). because js stopes working/js doesn't support multiple comment fields in one page .

### Use Case
1) loading comments of list of post objects in one page with comments. 
### Changes
1) only one line (I found out more code is needed) of js code need to be changed(in jquery version).
in no jquery version > few more lines.
I can create a pull request.

Radi85:
Hi @ChathuraGH,
Many thanks for your contribution in advance :wink: 

We would love to see your PRs.

ChathuraGH:
Hello @Radi85 👈. What's the problem here? 🤔.

ChathuraGH:
I thought this was easy.😲. Need more work. But possible.

PR #266, opened 2023-12-15
feat(#263): multiple comment fields in one page

ChathuraGH:
#263 feat
adding code for getting form's parent '#comments' for supporting multiple comment fields in one page.
at > 
1) submitDeleteCommentForm
2) submitCommentCreateForm

if we use jquery this is possible with one line.
`$form.parents("section#comments").replaceWith(data);`
It can be used in old versions that uses jquery. It is working 100%.

_without this code when you try to post or delete comment while having more than one comment fields in one page, the posting comments(parent comments) from all the comment fiels get add to frist comment field, When deleting all the deleted fields get updated to first field._

### DIFF
--- a/comment/static/js/comment.js
+++ b/comment/static/js/comment.js
@@ -213,7 +213,6 @@ document.addEventListener('DOMContentLoaded', () => {
                 // get parent element of the form with id=comments and replace html
                 var p = form.parentElement;while(p.id !== "comments"){var o = p;p = o.parentNode;};
                 p.outerHTML = result.data;
-                
             } else {
                 // child comment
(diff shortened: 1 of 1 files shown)


ChathuraGH:
Test the code before merging. I am not 100% sure it works. Even though it work on my pc.

ChathuraGH:
Pull request is incomplete. Because these changes are not enough to support multiple comment fields in one page.

Because, **Pagination** doesn't work. If one pagination link is clicked, all the comments fields in all the post objects are paginated. Not only clicked pagination comment field.

Expected behaviour  > 

Only clicked (pagination link) comment field should be loaded with corresponding comments.

